PR #9, opened 2016-01-06
Add bootstrap for stylus

simoneduca:


rogerhutchings:
Nice one!
- Files in the `global` dir need to be explicitly loaded in the correct order (`variables`, then `bootstrap`) so any custom variables override the defaults
- You can probably trash `global/box-sizing.styl`, as Bootstrap does this anyway
- Can you go through `global/bootstrap.styl` and be a bit more selective about what we're importing: we don't need `code`, or most of the components section. If you want to use glyphicons, you need to add the font files to the project somehow

rogerhutchings:
LGTM

shaunanoordin:
Review: Looks Good

Tested on:
- Chrome47/FF43/Edge25/IE11 + Win10
- Safari9/Opera34 + OSX

Comments:
- No issues detected across browser tests.
- Code passes sanity check.

Action: Merging

PR #10, opened 2016-01-07
Global CSS for Site Layout - Header, Footer, Content

shaunanoordin:
Layout functions as follows:
- header stays at the top of page.
- footer sticks to the bottom of the window or the bottom of the page, whichever has more height.
- content expands to fit the window, if it's smaller than the window.

Tested on:
- Chrome47/FF43/Edge25/IE11 + Win10
- Safari9/Opera34 + OSX


src/styles/global/ie11-fixes.styl
```diff
@@ -0,0 +1,6 @@
+html
+body
+#app-container
+#app-container > div
+  height 100%      //Enables display: flex child items to grow properly in IE11.
+  min-height 100%  //Prevents the above fix from limiting the height of said elements in other browsers.
```
rogerhutchings: How are these applied to IE11, using Modernizr or similar?

src/styles/components/site-footer.styl
```diff
@@ -0,0 +1,8 @@
+.site-footer
+  flex 0 0 auto  //Layout: Sticky footer. Sticks to bottom of window or bottom of page, whichever is larger.
+  
+  //The following isn't important and can be discarded.
```
rogerhutchings: If they can be discarded, they should be!

src/styles/global/ie11-fixes.styl
simoneduca: This conflicts with getting rid of `body.styl`. Can we put `app-container` somewhere else?

src/components/MapExplorer.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+export default React.createClass({
+  render() {
+    return (
+      <div id="map-explorer">
```
rogerhutchings: I'd prefer using classes to IDs for consistency. I think we only use `#app-container` to target the root element for react efficiently

src/styles/components/site-footer.styl
shaunanoordin: Ah, the "can be discarded" bits are just there to style the placeholder footer; they can be replaced with anything or removed entirely (although the footer would look ugly as heck). I only made that note because the only really important bit of CSS, layout-wise, is `flex: 0 0 auto`; every other bit of CSS is just deco so we can see _a_ footer.

src/components/MapExplorer.jsx
shaunanoordin: Affirmative on that; IDs will be swapped out with classes where possible.

src/styles/global/ie11-fixes.styl
shaunanoordin: @rogerhutchings These styles are applied normally across all browsers; (no Modernizr, etc) the fixes were separated into an independent ie11-fixes.styl file only for our labelling purposes.

@simoneduca Let me check in with you regarding removing body.styl; it's likely that I need to re-align myself with how everyone's planning to organise the CSS files.

simoneduca: @shaunanoordin sorry, poor communication on my part. I'll push the latest I've got to add-bootstrap PR, so you can have a look. Btw bootstrap should already take care of what you're describing.

shaunanoordin:
## Streamlined Layout CSS into a single global/layout.styl

Of note:
- app-container.styl, content-section.styl have been deleted since they contain no non-layout CSS
- IE11-fixes.styl has been rolled into global/layout.styl
- Main gist of the code that gets the layout working: #app-container (formerly #content) and .content-section are flex; header and footers are flex: 0

Not related to layouts, but still:
- Elements that use IDs (notably, #map-explorer) now use classNames instead, to keep things standardised.

Action: @simoneduca , @rogerhutchings , please review when convenient.

simoneduca:
Looks good to me. Happy to merge if @rogerhutchings gives ok.

rogerhutchings:
My only reservation is the sticky footer. In Annotate / SW, I ended up creating a footer that hides a bunch of stuff on the transcription interface, as the default Zoo footer is actually pretty huge. With this method, the footer would take up a lot of the map area.

I'm not saying we can't also do this in React - you'd need something to watch the active route, and toggle the links, blurb and whatnot off when you're on the map view.

Alternatively, we have some JS to watch the size of the viewport, modify the height of the component to equal that minus the header, and add the footer on afterwards.

**tldr**: I'm happy with this, as long as you've got a plan for how to incorporate the footer

shaunanoordin:
Good catch on that, @rogerhutchings ! Yes, I've seen what happens when the Zoo's mega-footer starts eating up into the main content space. I've simulated the problem by setting `.site-footer { height: 400px }` and we see this with our current code...

wildcam-education-20160107a

wildcam-education-20160107b

The .site-footer will squish the .content-section as much as it can, but it'll never squeeze it smaller than the available content - so our CSS is solid for long info pages. Yay!

For the more "app-like" pages such as the Maps, I see two solutions:
1. "shrink" the footer when React detects the user is on an "app-like" page, as you suggested, or
2. Set a min-height, e.g. `.content-section .map-explorer { min-height: 400px; }`, which is my current, functional solution

wildcam-education-20160107c

Frankly, I'd rather go with the more elegant Solution 1, so I'll explore React options for this.

I'm tentatively marking down a possible Solution 3, a CSS fix - 

```
@media screen and (min-height: 400px) {
  .site-footer .mega { display: none; }
  .site-footer .mini { display: block; }
}
```
- but I'll save that for another time; React gets my attention for now.

PR #14, opened 2016-01-15
Add table component. Add sidebar.

simoneduca:


simoneduca:
Thanks for the comments @rogerhutchings Fair point, I was just playing but it doesn't seem worth using the plugin. I'll try to use standard markup

shaunanoordin:
I managed to try out this branch, after temporarily removing Lines 34-36 of Sidebar.jsx. A few queries:

@simoneduca , is the Classroom Sidebar meant to be a permanent fixture (i.e. on every page) on the website?
- I can see situations where pages like the Map Explorer and any About pages may not require a Classroom Sidebar. (Although I can see a point keeping a record of the Classroom data in a global state.)
- If yes, this is a permanent fixture, then are there options to shrink/expand it? Similar to the footer?

@rogerhutchings , you mentioned that the Sidebar extends below the footer - where did you see this?
- I cannot replicate the Sidebar going below the footer - inspecting the element shows that the Sidebar element is about ~60px in height.
- I do, however, see that section.content-section now juts out to the right - a result of position:relative plus left:250px plus an implicit width 100%. Is this what you're talking about?
- (Verified on Chrome/FF/Safari on OSX.)

On that note: @simoneduca , can I suggest restructuring the HTML a bit, to prevent the section.content-section jutting out to the right?

Currently:

```
<div style="flex column">
  <header />
  <Sidebar />
  <section.section-content style="position: relative; left: 250px;">
    <!--child content-->
  </section>
  <footer />
</div>
```

If Classroom Sidebar is a permanent fixture on every page on the website

```
<div style="flex column">
  <header />
  <div style="display: flex; flex-direction:row;">
    <Sidebar style="flex: 1 1 auto" />
    <section.section-content style="flex: 4 4 auto">
      <!--child content-->
    </section>
  <footer />
</div>
```

If Classroom Sidebar is only specific to certain pages of the website

```
<div style="flex column">
  <header>
   <section.section-content>
    <!--child content start-->
    <div.content style="display: flex; flex-direction:row;">
      <Sidebar />
      <Stuff />
    </div>
    <!--end child content-->
  </section>
  <footer>
</div>
```

rogerhutchings:
@shaunanoordin I tried out the component with test data from the docs, which pushed the element length a lot longer than 60px. Give it a go and see what happens with some static html - it may be that the render doesn't trigger a redraw of the parent?

src/components/Sidebar.jsx
```diff
@@ -0,0 +1,41 @@
...
+        name: 'Classroom3'
+      },
+      {
+        name: 'Classroom4'
+      }
+    ]
+    });
+  }
+  render() {
+    let listItems = this.state.classrooms.map(function(classroom, i){
```
rogerhutchings: Let's do it the ES6 way:

```
let listItems = this.state.classrooms.map((classroom, i) =>
      <li key={i}>{classroom.name}</li>);

    return (
     <div className="sidebar">
        <h2>Classrooms</h2>
        <ul>
          {[...listItems]}
        </ul>
    </div>
```

src/components/Sidebar.jsx
```diff
@@ -0,0 +1,41 @@
...
+      return (
+        <li key={i}>{classroom.name}</li>
+      )
+
+    });
+    return(
+     <div className="sidebar">
```
rogerhutchings: fix indent

src/components/StudentTable.jsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
+
+export default React.createClass({
+  getInitialState() {
+      return {};
+    },
+  componentWillMount() {
+      this.setState({
+      students: [
```
rogerhutchings: fix indent - do you have brackets configured to use editorconfig?

src/components/StudentTable.jsx
```diff
@@ -0,0 +1,84 @@
...
+            { rows }
+          </tbody>
+        </table>
+      </div>
+    );
+
+  }
+});
+
+
```
rogerhutchings, line 81: only leave one trailing line

src/components/StudentTable.jsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
+
+export default React.createClass({
+  getInitialState() {
+      return {};
```
rogerhutchings: fix indent

src/components/Overview.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+export default class Overview extends React.Component{
+  constructor() {
+  super();
```
rogerhutchings: fix indent

src/components/App.jsx
```diff
@@ -9,18 +10,23 @@ export default React.createClass({
...
           <Link to="/about" className="link">About</Link>
           <Link to="/poweredby" className="link">Powered by</Link>
-          <Link to="/map" className="link">Map</Link>
         </header>
-        <section className="content-section">
-          {this.props.children || 'Welcome to React Starterify'}
-        </section>
+        <div className="container">
+          <div className="row">
+            <Sidebar/>
```
rogerhutchings: If the sidebar is specific to the admin area, it shouldn't be global (i.e. should be in the admin component)

src/components/StudentTable.jsx
simoneduca: I do. Mh...

PR #16, opened 2016-01-20
Bring project up to date with base

rogerhutchings:
Add Webpack, basically


rogerhutchings:
Ready for review @shaunanoordin @simoneduca

rogerhutchings:
This now also contains a deploy task and doesn't break the cameras view, go me :facepunch:

shaunanoordin:
Phew! I finally got this to work after a bit of a struggle.
- Code passes sanity check
- ~~Problem~~: Errors encountered when switching from `map-explorer-selectors` branch to `add-webpack` branch and then running `npm install` then `npm start` - website wouldn't compile.
  - Analysis: Error traced to conflicting versions of Babel on the two branches. (`npm install` doesn't detect when the correct dependency is installed, but it's the wrong version?) 
  - Actions: after switching to `add-webpack`, deleted /node_module and ran `npm install` again. `npm start` now works correctly.
  - Notes: Similar steps need to be taken when switching back to another branch with older dependencies, e.g. `map-explorer-selectors`
  - Status: Resolved

PR Status: **Validated**
Action: **Merging**

PR #21, opened 2016-01-29
Add authentication

simoneduca:


rogerhutchings:
The client has literally just been bumped to 2.0, and there are some other bits you need to make it work: see https://github.com/zooniverse/zoo-react-starterify/pull/36

PR #22, opened 2016-02-02
Add async

simoneduca:


shaunanoordin:
**Issue**: App would not load, citing the error: `Uncaught SugarClient.host is not defined`

**Analysis**: This issue stems from a bug within the Panoptes JS Client. In the client...

```
//Filename: lib/config.js
//...
var envFromBrowser = locationMatch(/\W?env=(\w+)/);
var envFromShell = process.env.NODE_ENV;
var env = envFromBrowser || envFromShell || DEFAULT_ENV;

module.exports = {
  host: hostFromBrowser || hostFromShell || API_HOSTS[env],
  clientAppID: appFromBrowser || appFromShell || API_APPLICATION_IDS[env],
  talkHost: talkFromBrowser || talkFromShell || TALK_HOSTS[env],
  sugarHost: sugarFromBrowser || sugarFromShell || SUGAR_HOSTS[env],
};
//...
```

...the `envFromShell` variable does NOT return a consistent value; sometimes the value is 'staging', sometimes it's 'development', etc, all depending on the Node shell running it. As a result, this will cause the `env` variable to take on an invalid value, causing things to go nuts when we try to have `API_HOSTS[invalid-env-value]`

**Actions**:
- Workaround: Instead of accessing localhost:3000, add a parameter that forces the Panoptes client to use a specific, valid environment value. Notably: http://localhost:3000?env=staging
- Solution: This issue needs to be patched at https://github.com/zooniverse/panoptes-javascript-client/

shaunanoordin:
Other than that issue, this PR looks good to me! Code passes sanity check and the App works fine with the workaround.

shaunanoordin:
WAIT! Sorry, sorry, I should've seen this earlier with a more thorough test. I caught something that's throwing an error on Chrome's console.

**Issue**: Accessing Admin > Students creates a `Reference Error: $ is not defined` error. This has a snowball effect that makes it impossible to navigate to other tabs/pages.
**Analysis**: The error seems to come from...

```
class StudentsContainer extends React.Component{
  createClassroom(name) {
    $.ajax.post({
      url: 'http://localhost:3000/teachers/classrooms',
      body: JSON.stringify({data: {attributes: {name: name}}}),
      headers: {'Authorization': 'Bearer ' + 'bcfb1211852702e3cfd745c6b557351e60f406d472bffe2803c93d79e6a05124'}
    })
  }
  ...
}
```

Tested on Chrome 48 on OSX10.11.2 with a standard node server. Note that this is a light test - I've managed to consistently replicate the error on a single testing environment, but I haven't delved with, e.g. different browsers.
**Action**: Further investigation required.

rogerhutchings:
$.ajax looks like jQuery code...

simoneduca:
My bad sorry. The PR isn't ready for review and I should have made that clear. @shaunanoordin yes `$.ajax` id just jQuery and it'll be removed when I get Redux middleware in place.

index.html
```diff
@@ -1,8 +1,11 @@
+<!-- CAN DELETE THIS -->
```
rogerhutchings: remove

index.html
```diff
@@ -1,8 +1,11 @@
...
 <!doctype html>
 <html>
   <head>
     <meta charset="utf-8">
     <title>Wildcam Gorongosa Education</title>
+    <!-- Import just Ajax -->
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/2.1.1/jquery.min.js"></script>
```
rogerhutchings: remove

src/actions/actions.js
```diff
@@ -0,0 +1,76 @@
...
+
+    dispatch(requestClassrooms(user))
+
+    // The function called by the thunk middleware can return a value,
+    // that is passed on as the return value of the dispatch method.
+
+    // In this case, we return a promise to wait for.
+    // This is not required by thunk middleware, but it is convenient for us.
+
+    return fetch('http://localhost:3000/teachers/classrooms')
```
rogerhutchings: shouldn't be hardcoded, derive the root from a config file

src/components/App.jsx
```diff
@@ -12,3 +12,4 @@ function mapStateToProps(state) {
   return { project: state.project }
 }
 export default connect(mapStateToProps)(App);
+// title { this.props.project.title }
```
rogerhutchings: remove

src/components/Classrooms.jsx
```diff
@@ -0,0 +1,59 @@
...
+      }]
+    };
+  }
+  // Fetch classrooms from education-api and update state
+  componentDidMount() {
+    this.fetchClassrooms = fetch(this.props.source, {
+      method: 'GET',
+      mode: 'cors',
+      headers: new Headers({
+          'Authorization': 'Bearer ' + '64aa60e488a555e2ef8b7f9d1621ec7e430dcb8b8baa0d00b8460eca5660d188',
```
rogerhutchings: shouldn't be hardcoded, again, stick it in a config file

src/components/Classrooms.jsx
```diff
@@ -0,0 +1,59 @@
...
+          'Content-Type': 'application/json'
+      })
+    })
+    .then(function(res) {
```
rogerhutchings: You've used this pattern in `actions.js`, along with fat arrows to save binding context:

```
promise
    .then(response => response.json())
    .then(json => {
        console.log('componentDidMountSTATE: ', this.state);
        this.setState({data: json.data})
        console.log('componentDidMountSTATE: ', this.state)
    });
```

src/components/Sidebar.jsx
```diff
@@ -1,37 +1,13 @@
...
-    let listItems = this.state.classrooms.map((classroom, i) =>
-      <li key={i}>{classroom.name}</li>);
 
     return (
       <div className="admin-sidebar">
-        <h2>Classrooms</h2>
-        <ul>
-          {listItems}
-        </ul>
+        <Classrooms source="http://localhost:3001/teachers/classrooms.json" />
```
rogerhutchings: shouldn't be hardcoded

rogerhutchings:
Aside from the comments, just do a sweep for missing semicolons and extra whitespace, and it should be there

src/components/Classrooms.jsx
simoneduca: Of course!

PR #30, opened 2016-02-09
Add root store with middleware.

simoneduca:
This is the first of a series of PRs intended to make network requests (fetching classrooms, login, etc) to work with redux-thunk middleware. The point of using a middleware layer is that we can dispatch other things other than actions, e.g. functions, to delay the execution of an action. 
Here `applyMiddleware` makes sure that the middleware is available in the actions dispatch method.
Webpack's Hot Module Replacement makes sure modules (reducers in this case) are kept updated.

### DIFF
--- a/src/store/configureStore.js
+++ b/src/store/configureStore.js
@@ -1,18 +1,17 @@
 import { createStore, applyMiddleware } from 'redux';
 import thunkMiddleware from 'redux-thunk';
-//import rootReducer from '../reducers';
 import rootReducer from '../reducers/reducers';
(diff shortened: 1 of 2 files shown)


src/store/configureStore.js
```diff
@@ -0,0 +1,18 @@
+import { createStore, applyMiddleware } from 'redux';
+import thunkMiddleware from 'redux-thunk';
+//import rootReducer from '../reducers';
```
rogerhutchings: trash this if we don't need it?

simoneduca: @rogerhutchings updated.

PR #32, opened 2016-02-09
Divvy up components into types

rogerhutchings:


shaunanoordin:
- Code passes sanity check, of course.
- Moving of files doesn't break anything. We're good!

PR Status: **Validated**
Action: **Merging**

PR #33, opened 2016-02-10
Request classrooms with middleware.

simoneduca:


simoneduca:
@rogerhutchings ready for review. Note: classrooms are not displayed, but I'd like to do that in a different PR to keep 'em short and sweet.

src/components/container/Classrooms.jsx
```diff
@@ -15,25 +19,14 @@ export default class Classrooms extends React.Component {
...
-          'Authorization': `Bearer ${eduApi.authToken}`,
-          'Content-Type': 'application/json'
-      })
-    })
-    .then(response => response.json())
-    .then(json => this.setState({data: json.data}))
-  }
+    console.log('componentDidMount');
+    const { dispatch } = this.props;
+    console.log('THIS PROPS', this.props)
```
rogerhutchings: remove, semicolon

src/reducers/reducers.js
```diff
@@ -15,17 +15,18 @@ function selectedClassroom(state = {}, action ){
...
     case types.REQUEST_CLASSROOMS:
+      console.log('REDUCERS: REQUEST_CLASSROOMS')
       return Object.assign({}, state, {
         isFetching: true
       })
     case types.RECEIVE_CLASSROOMS:
+      console.log('REDUCERS: RECEIVE_CLASSROOMS')
       return Object.assign({}, state, {
         isFetching: false,
         items: action.classrooms
```
rogerhutchings: Given that there's only one store, should the property it's stored under be a bit more descriptive (like `classrooms`)?

src/components/container/Classrooms.jsx
simoneduca: @rogerhutchings I assume the comma is not a typo :trollface: ?

src/reducers/reducers.js
simoneduca: Not sure about this one @rogerhutchings . The context of `items` should be clear because it is within `allClassrooms`. So I don't think `items` is ambiguous. This is what I mean:
screen shot 2016-02-11 at 10 17 33

src/components/container/Classrooms.jsx
rogerhutchings: derp. probably should just trash the line if we don't need it?

simoneduca: Yep.

src/reducers/reducers.js
simoneduca: Actually this doesn't matter because I realised that the object structure is wrong so `items` will need replacing anyway.

shaunanoordin:
- Code passes sanity check,
- ...but certain new features cannot be tested.
  - Said features require a currently unavailable Education API.
  - As a result, tests will log the (non-stopping) error: `http://localhost:3001/teachers/classrooms.json Failed to load resource: net::ERR_CONNECTION_REFUSED`
- However, since (1) this is a known issue, (2) said error doesn't compromise the overall stability of the project and (3) this update is necessary to bridge development with the Education API, I'm calling this PR validated and ready to merge.

Tested on: Chrome48/FF43/Safari9 on OSX

PR Status: **Validated**
Action: **Merging**

PR #35, opened 2016-02-11
Add redux logger

rogerhutchings:
@simoneduca - just a simple one to help debug redux


simoneduca:
Ah nice, I meant to add it in! Thanks

PR #36, opened 2016-02-12
Display classroom list

simoneduca:


src/containers/Classrooms.jsx
```diff
@@ -9,12 +9,11 @@ class Classrooms extends React.Component {
   constructor() {
     super();
     this.state = {
```
rogerhutchings: we don't need any of this, we need to set initial props instead!

rogerhutchings: so we trash the constructor, and at the end, add

```
Classroom.defaultProps = {
  selectedClassroom: {},
  allClassrooms: {
    classrooms: [],
    isFetching: false
  }
};
```

simoneduca: I don't think we need this...

PR #39, opened 2016-02-14
Display classrooms

simoneduca:
Add handling messages for api requests.
Make classroom list clickable.


simoneduca:
This branch had been already merged, but Roger branched off it and added some more features.
Merging again into master.

PR #49, opened 2016-02-23
Map Explorer Selector Panels Mk1

shaunanoordin:

This is a cleaner version of the cartodb-sql-selectors branch which had issues
merging with master after a disastrous rebase. See #46 for the original Lovecraftian journey into the mouth of madness.
- This is a checkpoint PR to ensure we're all in sync and to prevent this branch from growing too big.
- Added: **Selector Panels**, which allow users to choose their filter options. Multiple Selector Panels are allowed at a time, each corresponding to a different query, but on the first page load, one default Selector Panel is created.
- Added: **Advanced Mode** for Selector Panels, allowing users to type in SQL and CSS queries into CartoDB directly. Powerful, but essentially useless to our intended audience of educators without extensive documentation.
- Added but hidden (WIP): **Guided Mode** for Selector Panels, allowing educators to more easily select filters of interest.
- Added: Custom user input functionality - clicking on a point on the map will inform user of the camera's details.
- Added: Multiple Map Layers - we have a Satellite view and a Terrain (height/elevation) view, with the option to add as many more as we want.
- **Major Note**: There is a notable change in the CartoDB implementation, as our map now uses a Leaflet+CartoDB implementation.
  - Why: We had issues with the CartoDB-only implementation not responding to custom user input (clicking on features)
  - Pros: A base Leaflet layer means we can more easily switch out the CartoDB layer with e.g. a custom-built map visualisation feature (or perhaps Google Maps?)
  - Cons: Our Satellite view layer is worse than the ones we had before, because using Leaflet means accessing 'free' Satellite view 'tile/base layers'. (The CartoDB-only implementation had access to the 'Nokia Satellite Day' map from Here.com, which appears to be paid for in other cases: https://developer.here.com/plans/api/consumer-mapping)

@simoneduca , @rogerhutchings , this PR is ready for re-review.


simoneduca:
LGTM. As @rogerhutchings said before, the only concern, that should be addressed in the future, is stylistic: we should aim at breaking the mammoth MapExplorer into smaller componenets with the separation containers/presentational in mind. Waiting for comments to be addressed and then happy to merge. @shaunanoordin

shaunanoordin:
@simoneduca 
- Bindings have been moved to the constructor
- However, all other choices (colour, etc), editing or removing refreshUI (depending on our discussion) will be changed in `mapexplorer-selectors-mk2`

shaunanoordin:
Sorry, but this went kaput due to a bad rebase, once again. Replaced by #54.

PR #54, opened 2016-02-23
Map Explorer Selector Panels Mk1a

shaunanoordin:

I have terrible luck with rebasing. Replaces #49, which was supposed to replace #46. I will stay clear from rebases in the future.


simoneduca:
LGTM, merging.

Issue #26, opened 2016-02-05
Teacher / student routes need to be authenticated


rogerhutchings:
If you're not logged in, you should be required to authenticate before accessing those routes. Related to #21

shaunanoordin:
Here are some thoughts about Authentication, in relation to WGE Issue #26:

1: When an unautheticated user tries to access the /teachers (or /students) page, they will instead go to a page that says: "To access our features as a Teacher, please login: [button]". This can be done, no worries.

2: When an authenticated user clicks the Logout button, she/he needs to be **booted back to the Home** page, is everyone in agreement with this?

3: What happens when an authenticated user has two windows open, but logs out of one? The correct behaviour is for the second window to boot the user back to the home page too (as the first window would)... but if this is the case, we run into the matter of needing an elegant way to **"check-every-second/every-render-if-user-is-logged"**. Any thoughts on this?

rogerhutchings:
If we store the user in the redux state object, and have a listener on it, then it'll work in any window?

shaunanoordin:
Thanks @rogerhutchings , looks like redux & listening for the state update is the way to go.

I created a functional branch where I made the Panoptes client check user state in index.jsx and allowed/disallowed paths depending on user-login-state, but the darn thing wouldn't give a flying fart when the user clicked logout.

Follow up question: I recall you experimented with a method of URL-redirecting a user using react-router? I'd like to know if you found a good implementation for, e.g. bootUserTo('/home').

PR #55, opened 2016-02-23
New classroom form

simoneduca:


simoneduca:
@shaunanoordin ready to review.

shaunanoordin:
Code looks legit and the functionality of the Add New Classroom form works! I only have a few queries before I merge:

Q1: Just to confirm, `react-router` was upgraded to v2.0.0 for the `browserHistory` functionality, right?
Q2: The (optional) 'Subject' field of a class, after being set during the Add New Classroom process, isn't displayed anywhere. I assume this is a minor functionality that's coming in a later feature branch?

Notes:
- Unintended behaviour can be created by logging in & going to the Classrooms page, then clicking the Logout... because the page still LOOKS like you're logged in, allowing you to access Classroom and Add New Classroom features. (Although any attempt to actually add a new classroom while not logged in will fail silently)
- However, this is NOT a blocking factor for this branch - this is more related to Issue #26: Authentication priorities.

simoneduca:
@shaunanoordin 1) `react-router` has been upgraded because some feats of the older API will stop being supported/are already deprecated, e.g. mixins, history.transitionTo, etc.
2) correct
Notes should be taken care of when routes are properly authenticated.

shaunanoordin:
- Code sanity check passed
- Add New Classroom functionality verified.

PR Status: **Validated**
Action: **Merging**

PR #58, opened 2016-02-25
Display join url

simoneduca:


src/constants/config.json
```diff
@@ -6,8 +6,9 @@
   },
   "eduAPI": {
     "authToken": "64aa60e488a555e2ef8b7f9d1621ec7e430dcb8b8baa0d00b8460eca5660d188",
-    "classrooms": "teachers/classrooms.json",
-    "root": "https://education.staging.zooniverse.org/"
+    "root": "https://education.staging.zooniverse.org/",
+    "students": "/students/classrooms/",
+    "teachers": "teachers/classrooms/"
```
shaunanoordin, line 11: Query: is there a reason that `/students/classrooms/` starts with a `/` but `teachers/classrooms/` doesn't?

shaunanoordin, line 11: Observation: the leading `/` causes the URL link to be like: `https://education.staging.zooniverse.org//students/classrooms/137/join` (Note the double `//`)

src/presentational/Classroom.jsx
```diff
@@ -16,7 +38,14 @@ export default class Classroom extends Component {
...
           <TabPanel>
-            Overview
+            <h3>Join Link</h3>
+            <p>Send the following URL to all the students you want to join this classroom.</p>
+            <p>Note: students need to login to Zooniverse.org to be able to join.</p>
+            <input className="form-control" type="text" value={this.state.url} readOnly/>
+            <CopyToClipboard text={this.state.url} onCopy={this.onCopy}>
+              <button className="btn btn-default">Copy to clipboard</button>
+            </CopyToClipboard>
+            {this.state.copied ? <span style={{color: 'red'}}>Copied :)</span> : null}
```
shaunanoordin, line 48: Observation: clicking the Copy To Clipboard button causes the URL to be saved to be clipboard correctly (woot!) but the red-coloured 'Copied :)' text doesn't appear. This does not break functionality, but may be worth looking to in a future update.

shaunanoordin:
- Code passes sanity check, with minor comments:
  - Leading `/` in student config may require attention in the future
- Copy To Clipboard functionality confirmed
  - The 'Copied :)' message indicating a successful Clipboard Copy doesn't show, but this is a minor presentational issue instead of a functional one.

Tested on Firefox44 on SurfacePro3+Win10

PR Status: **Validated**
Action: **Merging**

src/constants/config.json
simoneduca, line 11: Nope, my bad. Well spotted, will fix, thanks.

src/presentational/Classroom.jsx
simoneduca, line 48: Sure will look into it.

PR #62, opened 2016-02-26
Enable students to join classrooms

simoneduca:
Add join action.
Add config routes.
Add JoinClassroom and JoinButton components.


simoneduca:
This is ready to go. @eatyourgreens or @shaunanoordin would be great if you could have a look.
Especially regarding the effects of changing `NODE_ENV` var in the webpack config.

src/constants/config.json
```diff
@@ -5,10 +5,14 @@
     "dataLayerIndex": 1
   },
   "eduAPI": {
-    "authToken": "64aa60e488a555e2ef8b7f9d1621ec7e430dcb8b8baa0d00b8460eca5660d188",
+    "testing-only-authToken": "64aa60e488a555e2ef8b7f9d1621ec7e430dcb8b8baa0d00b8460eca5660d188",
```
shaunanoordin, line 8: I just noticed that eduAPI's `authToken` variable isn't referenced anywhere else in the app - is this for future use or am I missing something? (e.g. implicit in one of our Panoptes libs.)

simoneduca, line 8: You're right, we can get rid of that.

src/constants/config.json
```diff
@@ -5,10 +5,14 @@
...
     "root": "https://education.staging.zooniverse.org/",
     "students": "students/classrooms/",
     "teachers": "teachers/classrooms/"
   },
+  "routes": {
```
shaunanoordin, line 13: Query: is there a reason we have similar `root` and `students` variable/values under `config.eduAPI` and `config.routes`?

shaunanoordin:
LGTM! I only have minor comments in the config file, but they're non-blocking. I'm happy to merge.
- Code passes sanity checks with only minor comments about extra variables in config.json
- Functionality confirmed - students can access the Join URL.
- New student landing page confirmed.

PR Status: **Validated**
Action: **Merging**

src/constants/config.json
simoneduca, line 13: It's a coincidence that the routes are similar. But I wanted to make clear that `routes` is for internal, app use, while `eduAPI` is used for external, API use.

PR #69, opened 2016-03-03
WIP - Do not merge - Production setup

simoneduca:
Add deploy to production step.
Add production url.
Add AppId from producation oauth.


simoneduca:
After some digging, I've found out that the new production domain needs to be white-listed on production server for preventing CORS errors. Once that's done, login should work.

simoneduca:
Closing as we can normally deploying from master now.

PR #76, opened 2016-03-07
Display classroom members.

simoneduca:
TODO: atm the same list of students is displayed for any classroom. FIX IT!

### DIFF
--- a/src/containers/Classroom.jsx
+++ b/src/containers/Classroom.jsx
@@ -8,10 +8,10 @@ import { default as ClassroomPresentational } from '../presentational/Classroom.
 export default class Classroom extends Component {
 
   render() {
-    const included = this.props.classrooms.included;
+    const members = this.props.classrooms.members;
     const classroom = this.props.classrooms.data.find(classroom =>
       classroom.id === this.props.params.classroomId);
-    return (<ClassroomPresentational data={classroom} included={included} />);
+    return (<ClassroomPresentational data={classroom} members={members} />);
   }
 
 }
@@ -25,7 +25,7 @@ Classroom.defaultProps = {
     data: [],
     loading: false,
     error: false,
-    included: []
+    members: []
   }
 };
 
--- a/src/containers/Classroom.jsx
+++ b/src/containers/Classroom.jsx
@@ -8,9 +8,10 @@ import { default as ClassroomPresentational } from '../presentational/Classroom.
 export default class Classroom extends Component {
 
   render() {
+    const included = this.props.classrooms.included;
     const classroom = this.props.classrooms.data.find(classroom =>
       classroom.id === this.props.params.classroomId);
-    return (<ClassroomPresentational data={classroom} />);
+    return (<ClassroomPresentational data={classroom} included={included} />);
   }
 
 }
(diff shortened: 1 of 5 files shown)


simoneduca:
@eatyourgreens or @shaunanoordin this is ready for review, thanks.

shaunanoordin:
Hey @simoneduca ! I've encountered a bit of wonkiness:

Issue: after loading a list of classrooms, clicking on a single classroom causes the whole app to 'crash', freezing in place and no longer responding to any user input.

Analysis: Error message
`Uncaught TypeError: Cannot read property 'length' of undefined` (Chrome)
`TypeError: allMembers is undefined` (FF)
at Classroom.jsx,

```
renderStudentList(allMembers, classroomMembers) {
    const list = (allMembers.length > 0) ? allMembers : 'No student found';
```

Hypothesis: if you haven't encountered this error on your end, then this may be specific to certain Zooniverse user accounts; perhaps the fact that I haven't joined any classes causes my allMembers list to be undefined?

Tested on Chrome48/FF44 on MacbookPro+OSX with Privacy Mode.

src/containers/Classroom.jsx
```diff
@@ -8,9 +8,10 @@ import { default as ClassroomPresentational } from '../presentational/Classroom.
 export default class Classroom extends Component {
 
   render() {
+    const included = this.props.classrooms.included;
     const classroom = this.props.classrooms.data.find(classroom =>
       classroom.id === this.props.params.classroomId);
-    return (<ClassroomPresentational data={classroom} />);
+    return (<ClassroomPresentational data={classroom} included={included} />);
```
shaunanoordin: Minor comment: the code in your update passes the sanity check very well, but the `included` variable isn't quite descriptive of its purpose. "What is `included`?" doesn't seem too clear to me; perhaps a brief //comment would help clarify it?

simoneduca: Ah yes, good catch!

src/presentational/Classroom.jsx
```diff
@@ -23,15 +24,43 @@ export default class Classroom extends Component {
...
-    let token = nextProps.data.attributes.join_token;
+    let classrooms = nextProps.data;
     this.state = {
-      url: routes.root + routes.students + 'join?id=' + id + '&token=' + token
+      url: routes.root + routes.students + 'join?id=' + classrooms.id + '&token=' + classrooms.attributes.join_token
     };
   }
 
+  renderStudentList(allMembers, classroomMembers) {
+    const list = (allMembers.length > 0) ? allMembers : 'No student found';
```
eatyourgreens: This sets `list` to a string if there are no students, which will break any code that depends on it being an array. Maybe set it `[]` instead?

simoneduca: Ouch! Good one, thanks! You're right this would break in the case there are no members at all yet.

simoneduca:
Review comments/issues should be sorted now. @shaunanoordin can you confirm you don't get the error anymore?

simoneduca:
Right, should be ready to go now @shaunanoordin

shaunanoordin:
Sweet!
- Previous issue addressed; error no longer detected.
- Teacher->Classroom pages are accessible as usual.
- Code passes sanity check.

Status: **Verified**
Action: **Merging**

PR #80, opened 2016-03-09
Map Explorer - Export Data as CSV Functionality

shaunanoordin:
Added: Export Data as CSV function in Map Explorer
- When user is in the Map Explorer, she/he can click on the "Update Map and Prepare CSV" button near the bottom of the left-hand selector panel.
- The feature allows users to download "live" data from CartoDB (i.e., as of writing, as recent as Jan 2016), based on the filter selections the user currently has activated, in the the form of (of course) a CSV.

Notes:
- We've **purposely locked the number of results** in the CSV to **1000 items**.
- The steps of the process is as follows:
  - User selects her/his filter options as normal.
  - User clicks on 'Prepare CSV' and is told to wait as the CartoDB SQL API is queried.
  - When the SQL data is returned, it's converted from JSON to CSV.
  - When that's complete, the "Download CSV" button appears; clicking it will prompt the user's browser to treat it like a file to download.
- A "screen-covering" popup dialog box is created during the Prepare CSV process to indicate to the user what's happening at any given time. Clicking outside the dialog box closes it. Opinion requested: does this seem intuitive enough?
  *I'm starting to consider spinning off this dialog box into its own component...

Ready for review, @simoneduca and @eatyourgreens .


simoneduca:
LGTM! Tested on latest v of FF, Safari and Chrome.
Non-blocking comment: perhaps it's more user-friendly to close the download modal on clicking the button, instead of needing to click outside of it. @shaunanoordin what do you think?

simoneduca:
Also, it looks like IE surprisingly supports `createObjectURL`. Can you confirm that? And how does it look on Edge?

shaunanoordin:
Thanks for the feedback, @simoneduca ! Agreed - on my next update of the Map Explorer, I'll make the Download modal close after the user has clicked the Download button.

I haven't tested on IE11 or Edge yet, but I did confirm that `createObjectURL` was compatible: http://caniuse.com/#search=createObjectURL . I'll doublecheck with my PC later today.

simoneduca:
Cool. It'll be deployed shortly.

PR #86, opened 2016-03-10
Add classroom metadata.

### LINKED ISSUE #51: Add metadata to classroom
> Additional metadata fields like "subject", maybe "school" and "description"

From #5 via @aliburchard

simoneduca:
Fix #51. @shaunanoordin this is really quick, if you fancy it.

### DIFF
--- a/src/actions/classrooms.js
+++ b/src/actions/classrooms.js
@@ -8,11 +8,14 @@ import * as types from '../constants/actionTypes';
 
 // Action creators
 
-export function createClassroom(name) {
+export function createClassroom(name, subject, school, description) {
   return dispatch => {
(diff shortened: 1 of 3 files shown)


shaunanoordin:
LGTM!
- Code passes sanity check
- User can add new classes with the following data: name, subject, school, description
  - Previous classes without pre-set data fields display blank fields, as expected.
  - Adding nonsensical data such as subject: `alert('x')` does not break anything
- User can browse created classes, which correctly contain the aforementioned data.

Tested on Chrome 48/FF44 on MacbookPro+OSX.

Status: **Validated**
Action: **Merging**

PR #91, opened 2016-03-15
WIP: Add a classrooms context for ClassroomsOverview

eatyourgreens:


src/reducers/classrooms.js
```diff
@@ -1,6 +1,6 @@
 import * as types from '../constants/actionTypes';
 
-const intialState = { loading: false, data: [], error: false };
+const intialState = { loading: false, data: [], error: false, members: [] };
```
eatyourgreens: missing i in `initialState` here?

src/reducers/classrooms.js
```diff
@@ -1,6 +1,6 @@
...
 
 
 export function classrooms(state = intialState, action) {
```
eatyourgreens: ditto

PR #99, opened 2016-03-18
[WIP] Map Explorer - Popup Dialog Screens (plus Code Cleanup)

shaunanoordin:
- Spun off the Dialog Screen popup feature as its own component.
- Dialog Screen also has child variants.
- Cleaned up code.


shaunanoordin:
I did so much cleanup and edits here that this can no longer be considered the `mapexplorer-species-popups`. Let me spin this off into a new branch before it turn into a mega-blob of an update.

PR #102, opened 2016-03-18
Map Explorer - Dialog Screen spinoff, Download File fixes and Cleanup

shaunanoordin:
This turned into a bit of a large update.
- The Dialog Screen has now been spun out, from a feature in MapSelectors to its own component... with its own child classes.
- There's a new fix for the Download CSV feature.
  - Previously, PC users would download a file without a .csv extension. Using the new `browser-filesaver` package and its `saveAs` feature, we now can send a file to the user for download with a specific file name.
- A bunch of features have been temporarily **hidden** (look for elements with `className="hidden"`).
  - This is in response to user feedback, which stated that there were too many extremely technical options.
  - We will revisit how to package these features in a more user-friendly way.
  - Please do not actually remove these features; it may not be visible, but the code still factors them into the CartoDB SQL and CSS calculations.
- A lot of code has been cleaned up; unused variables have been thrown out the window, etc.


shaunanoordin:
Almost forgot!
- Clicking on the Download button also causes the modal screen to close, as per Simone's suggestion: https://github.com/zooniverse/wildcam-gorongosa-education/pull/80#issuecomment-194803745

shaunanoordin:
@simoneduca and @eatyourgreens , this is ready for review. Cheers!

simoneduca:
I have no problem with the code. However, MapExplorer components are completely Redux unaware. This results in very big components which are quite difficult to follow, debug, maintain. implementing the action-reducer pattern of Redux will help improving the code. @shaunanoordin will look into this issue next week. Also react modals will be added instead of custom ones.
So, merging with these provisos.

PR #114, opened 2016-03-23
Teacher registration form

simoneduca:
Add dropdown component.
Hook it up to teacher registration form.


simoneduca:
Closing this because we're going for free text form.

PR #125, opened 2016-03-23
Fix #121

### LINKED ISSUE #121: Add loading spinners on classroom overview stats

simoneduca:


simoneduca:
@eatyourgreens or @shaunanoordin should be a quick one, thanks.

src/presentational/ClassroomsOverview.jsx
```diff
@@ -16,7 +18,9 @@ const ClassroomsOverview = (props, context) => (
                 <i className="fa fa-institution fa-4x"></i>
               </div>
               <div className="col-xs-9 text-right">
+                {(context.classrooms.data.length > 0) ?
                 <h1>{context.classrooms.data.length}</h1>
```
shaunanoordin, line 22: @simoneduca , I need to check: if a new user accesses this page for the first time, won't `context.classrooms.data.length` and `context.classrooms.members.length` both be 0? If yes, wouldn't new users always see the spinners, until they create a class and have a student?

Is there another way to check if the values have been properly loaded or not, perhaps `context.classrooms.loading === false`?

simoneduca, line 22: Derp, you're right! Fancy adding the fix?

shaunanoordin, line 22: Sure can do! Give me 5 mins.

shaunanoordin:
Awright, we're updated! :) I'm happy to continue merging if you have no issues, @simoneduca

shaunanoordin:
Status: **Works!**
Actions: **Merging**

Tested with a virgin Zoo account (no classrooms, no students) as well as a Zoo account that has classrooms and students, on Chrome49+MacbookPro+OSX.

PR #128, opened 2016-03-24
Collect teacher metadata

simoneduca:
Add teacher form with free text.
Create actions and reducers for users.


simoneduca:
@shaunanoordin or @eatyourgreens This is ready for review.
Expected behaviour:
If you are an educator and you haven't filled a registration form, you will see a "Register" button in the sidebar. Once you have filled in and submitted the form you should get redirected to the new classroom form. In future sessions, you will not be presented with the registration form again.

shaunanoordin:
Sorry @simoneduca , but the problem's still persisting. Good news is that it's not as bad as before, though. Here's where we stand now:

Issue: When a "virgin account" clicks on the new Register button, the app correctly opens the form for the user to fill in. However, when the user submits the form, the Classroom Sidebar - where the Register button and Create New Classroom button share a home - goes into perpetual Spinner mode.

screen shot 2016-03-24 at 17 32 28

Analysis:
- Navigating away from the Teachers page and back will fix the problem
- There are no error messages noted in the console, but check this out:
  screen shot 2016-03-24 at 17 43 29

So here's what happens...
1. On RECEIVE_USER, we get...

```
state.users {
  data: {
    attributes: {
      metadata: null
      zooniverse_display_name: "zootester6"
      zooniverse_id: "1325805"
      zooniverse_login: "zootester6"
    }
    id: "17"
    type: "users"
  }
}
```
1. However, on both UPSERT_TEACHER_METADATA and UPSERT_TEACHER_METADATA_SUCCESS we get...

```
state.users {
  data: {
    age: ""
    country: "Wop Bop a loo Bop"
    course: ""
    feedback: ""
    foundon: ""
    resources: ""
    setting: ""
  }
}
```

So it seems that UPSERT_TEACHER_METADATA/UPSERT_TEACHER_METADATA_SUCCESS incorrectly replaces state.users.data with the submitted form data. This explains why the Spinner continues to spin: `(!this.props.userdetails.loading && this.props.userdetails.data.attributes)` on `ClassroomSidebar` will never be `true` because `data.attributes` will be undefined.

Similarly, navigating away from the page and returning _will_ fix everything, because the `RECEIVE_CLASSROOMS` action will trigger, causing `state.users.data` to be correctly refreshed.

Tested on Chrome49+MacbookPro+OSX with two freshly created accounts.

shaunanoordin:
- Going into the Educators page using a new Panoptes account will prompt the user to Register themselves.
- After submitting the Registration form, the user is allowed to create New classrooms
- Extant accounts re-accessing the Educators page do not see the Register button again, but can straight away access the create New classroom funcitonality.
- All works as intended.

Tested on Chrome49+MacbookPro+OSX with two different Panoptes user accounts, one freshly created.

Status: **LGTM**
Actions: **Merging**

PR #131, opened 2016-03-24
MapExplorer / CartoDB now use Aggregations data instead of Classifications

shaunanoordin:
As @aliburchard pointed out, it'd make more sense for users to see data on the Map Explorer that's based on Aggregated Data, not raw Classifications Data, in most cases. (i.e. when viewing Global/World 'classifications'.) As a result:
- I've created [wildcam_gorongosa_aggregations_201603](https://shaunanoordin-zooniverse.cartodb.com/tables/wildcam_gorongosa_aggregations_201603) on CartoDB, based on recent data Greg compiled for me. (Thanks Greg!)
  - For the time being, that data is _static_. We'll wire up a live update pipeline in the near future, but for the time being we can assume that it's OK that aggregated data won't change as often as Classifications. (This is different from the time-sensitive Classifications data, where the whole idea is to see your individual contributions from recent classifications.)
- The SQL queries in Map Explorer have been changed to query the Aggregations table instead of the Classifications table.
  - The old SQL queries to Classifications still exist in the code, and will be returned to active service once the Classifications upload/sync pipeline is ready for it.
- Also, I created a small mod in the "onMapLayerClick" logic that will help with a future update - individual camera views. Don't mind it for now.


shaunanoordin:
@simoneduca and @eatyourgreens , this is ready for review.

PR #133, opened 2016-03-25
Add 'View Camera' functionality to Map Explorer

shaunanoordin:
screen shot 2016-03-25 at 20 08 21

Fun times!
- Clicking on a camera point on the Map will now create a popup modal showing select images from the camera.
- Only 6 random images from the camera will be shown, because I still have to think of a feasible way to show 1000+ images without things getting into batty territory, like spinning off a gallery feature.
- The images shown respect any filters selected by the user. So, for example, you can use the filter controls to show only Lion photos.
  - _Of course, this depends on the Aggregated data that we have at the moment, so... it looks great most of the time, but selecting for only Zebra photos leads to some silly results._

Please note that this branch depends on #131 to be merged first.


src/presentational/DialogScreen-ViewCamera.jsx
```diff
@@ -0,0 +1,30 @@
...
+      <section className={(this.props.status === DialogScreen.DIALOG_IDLE) ? 'dialog-screen' : 'dialog-screen view-camera enabled' } onClick={this.closeMe}>
+        <div className="dialog-box" onClick={this.noAction}>
+          {(this.props.message && this.props.message.length > 0)
+          ? <div>{this.props.message}</div>
+          : null}
+          
+          {(this.props.data)
+          ? <ul>
+              {this.props.data.map((item, index) => {
+                console.log(item);
```
eatyourgreens: Leftover debugging log?

eatyourgreens:
Trying this out now. I've tried clicking some of the grey circles on the map, but each one gives me a message like `There are no photos from camera E102`. Also, it's not immediately obvious what I should be clicking on the map. Can the cursor change over clickable objects?

eatyourgreens:
One other thing: I think this needs a `git rebase master` to avoid duplicating commits from #131

eatyourgreens:
Aha! Figured it out: the orange coloured circles are cameras with data. OK, this looks good enough for a demo to HHMI.

Couple of points re. the dialog. It's only operable with the mouse at the moment. At the very least, pressing `esc` should close it. I'd suggest:
- adding `role=dialog` to the containing `div`
- focusing the dialog when it opens, and not allowing it to lose focus while open. Modals should basically steal keyboard focus from the underlying page and not give it back until they close.
- listening for `keydown` or `keypress` on the element with `role=dialog` and closing if the key is `esc`

but these changes can be made in a separate PR, rather than holding up this one.

shaunanoordin:
Thanks for your feedback, @eatyourgreens and @aliburchard ! 

I've cleaned up the PR with the following:
- branch rebased to the latest master.
- Added a Map Legend that helps explain the black circles and orange circles, and the fact that you can click them.
- Cleaned up the Dialog Screens a bit with better styling and added `role=dialog`.

I'm going to spin off the following into a separate dev branch:
- DialogScreens needs to steal focus when they're active. (as per Jim's suggestion)
- Review the use of keyboard input in MapExplorer; current incarnation is too mouse-centric. (Jim's)
- The Apply and Download buttons needs to always be visible on screen; otherwise, users may not be aware that they need to click on them to change things. (Ali's)
- "Empty" (black) cameras need to be replaced with an `X` symbol, or similar, to clearly indicate that there are no cameras. (Ali's)
  - (Derp: simple solution is to set an SVG in the CartoCSS, but dangit, I can't seem to get it to work with a local resource - I might need to upload the SVG to production first.)
    `
    #wildcam_gorongosa_compiled_201601 [ count = 0] {
    marker-width: 10.0;
    marker-file: url(http://www.clipartbest.com/cliparts/nTX/AGX/nTXAGXyTB.svg);
    }
    `
- The Camera View should have more data (e.g. "You are viewing Camera C69, which is 13m from water and in a Miombo Woodland") (Ali's)
  - The corresponding downloaded CSV should have more aggregation data inside it. (Ali's)

Unfortunately I still have issues with:
- Getting markers to change styles (e.g. highlighting) on mouse over. Problem is, the collection of markers you see aren't individual DOM/SVG elements, but actually a single raster image with a bunch of orange/black dots. I'll need to think of an alternative way of indicating to the users that "yep you can click this circle", possibly using the `.on('featureOver', {})` event listener to... create a tooltip or something.
- Getting that Map Legend to work _without_ using innerHTML - I'll need to dig into the Leaflet documentation to figure out why it's incompatible with JSX tags. I mean it works, it's just that the code doesn't look neat.

@simoneduca and @eatyourgreens , ready for review! :grinning:

eatyourgreens:
:+1: Looks good to me.

PR #142, opened 2016-03-30
Fix #141

### LINKED ISSUE #141: Classroom list data is undefined when joining a classroom
screen shot 2016-03-30 at 13 47 37

Thanks @aliburchard for spotting this.

simoneduca:

### DIFF
--- a/src/actions/classrooms.js
+++ b/src/actions/classrooms.js
@@ -129,16 +129,3 @@ export function fetchStudentClassrooms() {
   }
 }
 
-
-//.then(json => {
-//      dispatch({
-//        type: types.JOIN_CLASSROOM_SUCCESS,
-//        loading: true
-//      });
-//      console.log('JOIN-RESPONSE: ', json.data);
-//      // the following reloads the page causing the classroomlist to refresh
-//      // not sure if there is a better way that uses react-router methods.
-//      window.location.assign('/students/classrooms/')
-//    })
-
-
--- a/src/actions/classrooms.js
+++ b/src/actions/classrooms.js
@@ -64,7 +64,6 @@ export function joinClassroom(id, token) {
     })
     .then(response => response.json())
     .then(json => {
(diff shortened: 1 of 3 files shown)


src/actions/classrooms.js
```diff
@@ -131,4 +130,15 @@ export function fetchStudentClassrooms() {
...
+//      dispatch({
+//        type: types.JOIN_CLASSROOM_SUCCESS,
+//        loading: true
+//      });
+//      console.log('JOIN-RESPONSE: ', json.data);
+//      // the following reloads the page causing the classroomlist to refresh
+//      // not sure if there is a better way that uses react-router methods.
+//      window.location.assign('/students/classrooms/')
+//    })
+
```
eatyourgreens: Delete all these commented lines?

simoneduca: AGH! Done, thanks.

shaunanoordin:
Good news is that Joining a Classroom properly succeeds in adding a student to the classroom via the API. Also, the previously encountered `can't find .length() of undefined` error is gone - the app is stable.

Bad news is that, after joining a Classroom, the list of Classrooms on the sidebar doesn't properly refresh to reflect my newly joined class.

For example:
- I am a student with existing classrooms Apple and Banana.
- I click a link to join classroom Cherry
- The join is successful, but the Classroom list only shows Apple and Banana
- Navigating away and returning to the page will, however, correctly show Apple, Banana and Cherry listed as my classrooms.

screen shot 2016-03-30 at 14 39 50

EDIT: As @eatyourgreens just pointed out, this may be due to the JOIN_CLASSROOM and RECEIVE_STUDENT_CLASSROOM actions being asynchronous, so the list-all-classes doesn't necessarily trigger _after_ the successful join action.

shaunanoordin:
Status: **OK** This branch is stable enough to merge and fixes the app, but we'll need to revisit this later to address the "not showing most recently joined classroom" issue.
Actions: **Merging**

PR #145, opened 2016-03-30
Fix #134

### LINKED ISSUE #134: Make all fields of teacher form required.

simoneduca:
This is the quickest way, but I've been working on a more robust way of doing it that we may include later.
@eatyourgreens @shaunanoordin this is ready to go.


shaunanoordin:
Simple and functional - the Registration Form doesn't allow empty fields to be submitted. (Didn't test with nonsensical data however, such as pure whitespace.) Tested on Chrome 49 with a new Zooniverse account.

Status: **LGTM**
Actions: **Merging**

PR #147, opened 2016-03-30
WIP: Update student classrooms after joining

eatyourgreens:


eatyourgreens:
Talking to @rogerhutchings I think what needs to happen is here is that we trigger a new action, which either completely refreshes the classroom list, or adds the new classroom to the existing list.

simoneduca:
@eatyourgreens that doesn't work for me. The application hangs and the student can't join and the list isn't updated.
screen shot 2016-03-30 at 16 47 49

simoneduca:
Superseded.

PR #148, opened 2016-03-30
Map Explorer - Improved SQL Queries & Results

shaunanoordin:
- CartoDB now updated with improved Aggregations Data (thanks Greg!) and more Subjects data (16000 items instead of 10000 previously)
- Map Explorer now uses improved SQL queries to pull more relevant data for users to view. No more 'nonsense' results, like Elephants being labelled as Zebras - those stripes aren't that slimming, damn it.
  - Yes, there is a variable called `sqlWhere_arbitraryMagicVoodoo`, and it's called that _specifically_ because we need to revisit this issue and ask whether data should be cleaned at runtime, or pre-processed. (You have an idea what I prefer.)
- As a side effect to the change in SQL queries, the Download CSV feature now features _all the aggregation data_, which is something I'm sure Ali is super happy with. Hooray for data!

@simoneduca and @eatyourgreens , please let me know when you can review this; I plan this to be my last PR for the 'public test launch' tomorrow.


eatyourgreens:
:+1:

PR #150, opened 2016-03-31
Update classrooms list after student joins.

simoneduca:


eatyourgreens:
:+1:

PR #152, opened 2016-03-31
Improve Map Explorer's Centre View on first load

### LINKED ISSUE #149: Improve Map Explorer's Centre View on first load...?
Goddammit, I just realised that with the new SQL removing 'crappy' results, viewing the Map Explorer on a smaller screen (e.g. MacBook Pro screen, as opposed to a larger monitor) will make the Map Explorer look like this on first load:

screen shot 2016-03-30 at 17 48 13

I can change the centre-position and zoom of the map when it first loads, so it'll look like this on a MacBook Pro...

screen shot 2016-03-30 at 17 47 49

...but then it gets a bit more off-centre on larger screens.

screen shot 2016-03-30 at 17 54 38 2

I already have a branch with this 'fix' - https://github.com/zooniverse/wildcam-gorongosa-education/tree/mapexplorer-beta-fixes - but the question is, do we want to fix this _now_? Since 1. this is technically just a patch over a non-essential thing right before a launch, and 2. no matter what, the users need to fiddle with the map to get their own "centre" position anyway, depending on their screen size? But 3. it does affect presentation and we do want to make a good first impression?

Thoughts, @eatyourgreens @aliburchard @simoneduca ?

shaunanoordin:
- Fixes #149 
- Basically, we're setting a new center position for the Map Explorer on first load.


eatyourgreens:
:+1: We can look later at dynamically setting the map centre, based on viewport size, if that's a problem.

PR #155, opened 2016-03-31
Remove action

simoneduca:
When I fixed the student classrooms issue I forgot to clean up some old code.
@shaunanoordin thanks.

### DIFF
--- a/src/containers/JoinClassroom.jsx
+++ b/src/containers/JoinClassroom.jsx
@@ -12,7 +12,6 @@ class JoinClassroom extends Component {
 
   join() {
     console.log('Joining ...')
-    this.props.dispatch(fetchStudentClassrooms());
     this.props.dispatch(joinClassroom(this.props.location.query.id, this.props.location.query.token));
 
   }


shaunanoordin:
Works great!

To be clear to everyone, the issue was this:

Issue: If Alice create Classroom X, and Bob clicks the Join button for Classroom X, then Bob's Student Sidebar will _very briefly_ show Classroom X before it disappears. Bob has actually joined Classroom X, but the Student Sidebar seems shy about admitting it.

Analysis: This was caused by two simultaneous asynchronous calls, which would (sometimes) cause the following to happen:
1. REQUEST_STUDENT_CLASSROOMS
2. JOIN_CLASSROOM
3. JOIN_CLASSROOM_SUCCESS, which properly shows Classroom X added to Bob's list of classes.
4. RECEIVE_STUDENT_CLASSROOMS, which proceeds to overwrite the update list of classes with the 'old version' where Classroom X isn't listed.
As the calls were **asynchronous**, it was difficult to test at a cursory glance. 

Solution: As Simone's fix shows, there was an unnecessary `fetchStudentClassrooms()` call which incorrectly triggered actions 1 and 4 above. For the `join()` action, only 2 and 3 were needed.

Status: **Verified working**
Actions: **Merging**

PR #156, opened 2016-03-31
Create env-specific config

rogerhutchings:
Should fix any issues with env specific deployment, but it needs to be tested on production after Ali's finished her workshop

@simoneduca


simoneduca:
Great thanks @rogerhutchings

simoneduca:
LGTM :+1:

PR #157, opened 2016-04-01
Fancier teacher registration form

simoneduca:


src/presentational/Checkbox.jsx
```diff
@@ -0,0 +1,47 @@
+import { Component, PropTypes } from 'react';
+
+export default class Checkbox extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      selected: 'selected',
```
eatyourgreens: why is the initial value 'selected'? shouldn't it be `true` or `false`?

src/presentational/Checkbox.jsx
```diff
@@ -0,0 +1,47 @@
...
+
+  handleChange(e) {
+    this.props.onChange(e);
+    this.setState({selected: e.target.checked});
+  }
+
+  render() {
+    const options = this.props.options.map((option) =>
+      <div>
+        <input
```
eatyourgreens: You should replace the `div` here with a `label` element, wrapping the input and its associated text label.

src/presentational/Checkbox.jsx
```diff
@@ -0,0 +1,47 @@
...
+          key={option.value}
+          name={this.props.name}
+          type="checkbox"
+          value={option.label}
+          onChange={this.handleChange} />
+            {option.label}
+        <br />
```
eatyourgreens: Delete this `br` and use CSS to lay out your list of checkboxes.

src/presentational/Checkbox.jsx
```diff
@@ -0,0 +1,47 @@
...
+      </div>
+    )
+    return (
+      <div>
```
eatyourgreens: I'd get rid of the nested `div` elements and mark this up as a `ul` list instead, with each input inside an `li`, then style the list with CSS to get either a vertical or inline layout.

src/presentational/Dropdown.jsx
```diff
@@ -0,0 +1,46 @@
+import { Component, PropTypes } from 'react';
+
+export default class Dropdown extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      selected: 'selected',
```
eatyourgreens: Initial state of selected should be `null` if nothing is selected?

src/presentational/Dropdown.jsx
```diff
@@ -0,0 +1,46 @@
...
+    )
+    return (
+      <select
+        className='form-control'
+        name={this.props.name}
+        value={this.state.selected}
+        onChange={this.handleChange}>
+          {options}
+      </select>
+    )
```
eatyourgreens, line 35: Needs a `label` and the text label for the dropdown box eg. 'Country' before the `select`.

src/presentational/TeacherForm.jsx
```diff
@@ -87,85 +93,82 @@ class TeacherForm extends Component {
...
-           <label>Where do you teach?</label>
-           <input className="form-control"
-            type="text"
-            name="country"
-            placeholder="Insert a country"
-            autofocus="true"
-            value={this.state.country}
-            required
-            onChange={this.handleChange}/>
+            <label>Where do you teach?</label>
```
eatyourgreens: This label isn't associated with the dropdown. It might be better to make it part of the Dropdown component. (see above.)

src/presentational/TeacherForm.jsx
```diff
@@ -87,85 +93,82 @@ class TeacherForm extends Component {
...
-           <label>In what educational setting do you plan to use this resource?</label>
-           <input className="form-control"
-            type="text"
-            name="setting"
-            placeholder="E.g. Classroom, home"
-            autofocus="true"
-            required
-            value={this.state.setting}
-            onChange={this.handleChange}/>
+            <label>In what educational setting do you plan to use this resource?</label>
```
eatyourgreens: See above. `label` elements need to be explicitly associated with a single form control. See the examples at https://developer.mozilla.org/en/docs/Web/HTML/Element/label

src/presentational/TeacherForm.jsx
```diff
@@ -87,85 +93,82 @@ class TeacherForm extends Component {
...
+            <Checkbox className="form-control"
```
eatyourgreens: Maybe call this component `CheckboxGroup` since it displays a list of checkboxes, not a single checkbox?

Also, each individual checkbox will require its own `label` but you can use `fieldset` to group several related controls, like these. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset

src/presentational/TeacherForm.jsx
```diff
@@ -87,85 +93,82 @@ class TeacherForm extends Component {
...
-           <label>I plan to use this resource with my students.</label>
-           <input className="form-control"
-            type="text"
-            name="age"
-            placeholder="E.g. Junior high, 2-yr College"
-            autofocus="true"
-            required
-            value={this.state.age}
-            onChange={this.handleChange}/>
+            <label>I plan to use this resource with my students.</label>
```
eatyourgreens: See above re. form labels

src/presentational/TeacherForm.jsx
```diff
@@ -87,85 +93,82 @@ class TeacherForm extends Component {
...
-           <label>In what course(s) do you plan to use this resource?</label>
-           <input className="form-control"
-            type="text"
-            name="course"
-            placeholder="E.g. Biology, Ecology"
-            autofocus="true"
-            required
-            value={this.state.course}
-            onChange={this.handleChange}/>
+            <label>In what course(s) do you plan to use this resource?</label>
```
eatyourgreens: Label needs to be associated with its form input.

src/presentational/TeacherForm.jsx
```diff
@@ -87,85 +93,82 @@ class TeacherForm extends Component {
...
-           <input className="form-control"
-            type="text"
-            name="foundon"
-            placeholder="E.g. Zooniverse, Social Media"
-            autofocus="true"
-            required
-            value={this.state.foundon}
-            onChange={this.handleChange}/>
+            <label>How did you find out about this resource?</label>
+            <input className="form-control"
```
eatyourgreens: Label needs to be associated with its form input.

src/presentational/TeacherForm.jsx
```diff
@@ -87,85 +93,82 @@ class TeacherForm extends Component {
...
-           <label>Have you used HHMI BioInteractive resources other than WildCam Gorongosa in your teaching before?</label>
-           <input className="form-control"
-            type="text"
-            name="resources"
-            placeholder="Yes or no"
-            autofocus="true"
-            required
-            value={this.state.resources}
-            onChange={this.handleChange}/>
+            <label>Have you used HHMI BioInteractive resources other than WildCam Gorongosa in your teaching before?</label>
```
eatyourgreens: Label needs to be associated with its form input.

src/presentational/TeacherForm.jsx
```diff
@@ -87,85 +93,82 @@ class TeacherForm extends Component {
...
-           <label>Feedback from educators like you helps us improve our free educational resources. May we contact you at a later time?</label>
-           <input className="form-control"
-            type="text"
-            name="feedback"
-            placeholder="Yes or no"
-            autofocus="true"
-            required
-            value={this.state.feedback}
-            onChange={this.handleChange}/>
+            <label>Feedback from educators like you helps us improve our free educational resources. May we contact you at a later time?</label>
```
eatyourgreens: Label needs to be associated with its form input.

simoneduca:

src/presentational/RadioButtonGroup.jsx
```diff
@@ -0,0 +1,52 @@
...
+    this.state = {
+      checked: null,
+    }
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  handleChange(e) {
+    this.props.onChange(e);
+    this.setState({checked: e.target.checked});
+  }
```
eatyourgreens, line 11: Does `state.checked` ever get used? I can't see it anywhere else in this component.

src/presentational/CheckboxGroup.jsx
```diff
@@ -0,0 +1,65 @@
...
+        values.push(options[i].value);
+      }
+    }
+    return values;
+  }
+
+  handleChange(e) {
+    e.target.value = this.getCheckboxValues(e.target.name);
+    this.props.onChange(e);
+    this.setState({checked: e.target.checked});
```
eatyourgreens: Does `state.checked` ever get used? I can't see it anywhere else in this component.

eatyourgreens:
Looks good to me, apart from those comments about `state.checked` for the checkboxes and radio buttons.

simoneduca:
@eatyourgreens thanks! I removed `state.checked`, I forgot it there from the Dropdown component. This should be good to go.

eatyourgreens:
:+1:

Small style point: from the name, I'd expect `checkArray` to return `true` if the form was valid so I had to double-check that bit to make sure the code was checking the form properly.

simoneduca:
@eatyourgreens thanks! I could change it to `return array.some(item => item !== null && item !== '')` if it makes it clearer?

eatyourgreens:
@simoneduca `return !array.some(item => item === null || item === '')` would flip the true/false value of the return value, but it's a minor point.

Your updated check will return `true` if some fields are filled out, but not all. I don't think that's what you want.

PR #159, opened 2016-04-04
Add deploy-production script

simoneduca:


shaunanoordin:
(Heya, just poking my head in here.) The code change LGTM, but I'm curious to know if there's a way to test this without making an actual live deploy. I guess no?

simoneduca:
@shaunanoordin that's a good question, I'm not sure...

PR #160, opened 2016-04-04
Login Optimisations - Bug Fixes and Redux Refactoring

### LINKED ISSUE #75: (Optimisation) Authentication/Login to be refactored
The Authentication/Login mechanics need to be optimised.
- Issue: Login logic works, but does not follow standard redux patterns.
- Goal: Login logic should work more naturally within flow of component lifecycles+redux stores.
- Suggestion: Move login logic, which currently sit in `container/HeaderAuth.jsx`, into the redux actions in `actions/login.js`

Follow up to merged PR #70.

Note the comments in that PR:
- _The Login/Logout logic currently sits in HeaderAuth, although a future improvement would be to move the logic into the Redux Actions themselves._
- _[Weird] The Logout action in HeaderAuth uses window.location to move logged out users to the main page. Strangely, if we use browserHistory.push, the HeaderAuth doesn't update the render elements properly (i.e. changing the Logout to Login). Any thoughts on this would be appreciated._

shaunanoordin:
It's Login, refactored and bug-fixed!
- All Login functions have been moved from the container/presentational components to actions/login.js. This has lead to a reduxion of complexity.
- When users log in, they stay logged in _properly_.
  - That's right! If you're looking at the Teachers page and refresh, you no longer get booted out!
  - The trick was to realise that there are actually _three states_ to the user login: user is logged in, user is logged out, and _we have no idea either way_ (i.e. status not initialised).
- Logout button doesn't do a horrible redirect like before. Now, if you logout on a login-secured page, you're directed back to the login prompt. if you logout on any other page, the app goes, "ehh, that's fine, carry on."
- A small "loading screen" appears if the user's first page is a login-secured page, as the app checks the user's login status.
- Code now properly rebased to take into account new `config.js` file, which intelligently picks between production and staging environments. That said, **if login goes wonky on production, please let me know**.

Closes #75

### DIFF
--- a/src/actions/login.js
+++ b/src/actions/login.js
@@ -1,12 +1,38 @@
 import * as types from '../constants/actionTypes';
+import Panoptes from 'panoptes-client';
+import config from '../constants/config';
 
-// Action creators
+//Action creators
+
+export function checkLoginUser() {  //First thing on app load - check if the user is logged in.
+  return (dispatch) => {
+    Panoptes.auth.checkCurrent()
+      .then((user) => {
+        dispatch(setLoginUser(user));
+      });
+  }
+}
 
 export function setLoginUser(user) {
-  return dispatch => {
+  return (dispatch) => {
     dispatch({
       type: types.SET_LOGIN_USER,
       user
     });
   };
+}
+
+export function loginToPanoptes() {  //Returns a login page URL for the user to navigate to.
+  return (dispatch) => {
+    return Panoptes.oauth.signIn(config.panoptesReturnUrl)
+  }
+}
+
+export function logoutFromPanoptes() {
+  return (dispatch) => {
+    Panoptes.oauth.signOut()
+      .then(user => {
+        dispatch(setLoginUser(user));
+      });
+  }
 }
--- a/src/constants/mapExplorer.config.json
+++ b/src/constants/mapExplorer.config.json
@@ -226,7 +226,7 @@
     {
       "id": "sable",
       "dbName": "Sable Antelope",
-      "displayName": "Sable Atelopes"
+      "displayName": "Sable Antelopes"
     },
     {
       "id": "samango",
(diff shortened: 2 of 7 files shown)


shaunanoordin:
Oops, forgot to add: @simoneduca and @eatyourgreens , this is ready for review. Woot!

eatyourgreens:
The Sable Antelopes typo looks a bit out-of-place here, but I'm happy to merge this.

simoneduca:
I would only add that this might be a good place/time to grab constants such as `panoptesRedirectUri` from the window object. I think it's a good idea because it'd mean one less config value to check. So something like `window.location.hostname`

eatyourgreens:
In that case, don't you potentially run into cases where the `redirectUri` may not match the allowed values that you've set for the app, so the auth request will be rejected?

simoneduca:
I think that would be a problem with `location.href` but not with `hostname`, but I'm happy to be proven wrong. Have you got an example in mind @eatyourgreens ?

simoneduca:
Anyway, I didn't mean to block the merge :smile:

eatyourgreens:
Does this app always redirect back to the home page after authorisation? In which case, ignore me.

eatyourgreens:
@simoneduca I think I was confusing this with oAuth issues on Shakespeare's World.

shaunanoordin:
Thanks for the feedback!

@simoneduca, for your suggestion, one possible implementation is to replace instances of
`Panoptes.oauth.signIn(config.panoptesReturnUrl)`
with
`Panoptes.oauth.signIn(window.location.origin || (window.location.protocol + window.location.hostname))`
_(Internet Explorer doesn't support window.location.origin, otherwise this would've been much cleaner.)_

That said, I'm OK with keeping things the way they are - explicitly setting a Return URL in the config - because a Return URL is also explicitly set in the Panoptes auth service.

shaunanoordin:
**Side note!**: a more advanced feature I was toying with was the ability to return users to the page they were trying to view before they logged in. e.g. if they went to the Teachers page, but were prompted to log in, then after they logged in they should have been _redirected back to_, e.g. `Panoptes.oauth.signIn("https://learn.wildcamgorongosa.org/teachers/classrooms")`

Unfortunately, the current Panoptes settings don't allow a return URL that's not the exact, specific one explicitly set by the app. Boo!

PR #162, opened 2016-04-05
Fix lack of PublicPath in Production's Webpack Config

shaunanoordin:
Issue: If you go directly to https://learn.wildcamgorongosa.org/teachers/classrooms, the app won't load.
Analysis: This is caused by the app trying to load <script src="main-8874ee12d8d7858711ca.min.js"></script> instead of <script src="**/**main-8874ee12d8d7858711ca.min.js"></script>.
Solution: Add `publicPath` to `webpack.production.config.js`.

### DIFF
--- a/webpack.production.config.js
+++ b/webpack.production.config.js
@@ -18,6 +18,7 @@ module.exports = {
   output: {
     path: path.join(__dirname, '/dist/'),
     filename: '[name]-[hash].min.js',
+    publicPath: '/',
   },
 
   plugins: [


shaunanoordin:
@simoneduca and @eatyourgreens , this is ready for review.

simoneduca:
As Shaun pointed out, we were already using this when deploying from the production-setup branch https://github.com/zooniverse/wildcam-gorongosa-education/blob/production-setup/webpack.production.config.js#L21

This might also fix #163 

LGTM.

PR #164, opened 2016-04-05
CartoDB SQL & CSS polish

shaunanoordin:
Minor tweaks to the CartoDB SQL and CSS
- The 'empty' black cameras are back, woo! They still need to be replaced with black Xs though.
- Magic voodoo nonsense removed.
  - NOTE: We'll need to revisit the issue of 'cleaning' the data further up the pipeline
- Also removed other bits of code that aren't used any more, e.g. vestigial map config data.

Ready for review, but this isn't a priority item, so no rush.


src/containers/MapExplorer-SelectorPanel.jsx
```diff
@@ -69,7 +69,7 @@ export default class SelectorPanel extends React.Component {
     return (
       <article className="selector-panel">
         <section className={(this.props.selectorData.mode !== SelectorData.GUIDED_MODE) ? 'input-subpanel not-selected' : 'input-subpanel' } ref="subPanel_guided">
-          <h1 onClick={this.changeToGuided}>Standard Mode</h1>
+          <h1 className="hidden" onClick={this.changeToGuided}>Standard Mode</h1>
```
eatyourgreens: Headings can't be focussed with the tab key, so this can only be used with a mouse or touch. Better to use a button for clickable controls like this.

src/containers/MapExplorer-SelectorPanel.jsx
```diff
@@ -120,7 +120,7 @@ export default class SelectorPanel extends React.Component {
           </div>
         </section>
         <section className={(this.props.selectorData.mode !== SelectorData.ADVANCED_MODE) ? 'input-subpanel not-selected' : 'input-subpanel' } ref="subPanel_advanced" >
-          <h1 onClick={this.changeToAdvanced}>Advanced Mode</h1>
+          <h1 className="hidden" onClick={this.changeToAdvanced}>Advanced Mode</h1>
```
eatyourgreens: Ditto.

eatyourgreens:
Looks good, apart from the clickable headings. Not sure why github is telling me I commented on an outdated diff.

shaunanoordin:
:ok_hand: Boop! I've made the clickable headings into buttons.

The headings will actually be hidden from the users at this stage of development, but you make a good point about the accessibility issues of a clickable non-buttons, which needs to be carried over as the map UI panels evolve.

PR #165, opened 2016-04-06
Fixes Issue 153: Student Members are now counted uniquely

### LINKED ISSUE #153: Students number includes students with the same zooniverse username
screen shot 2016-03-31 at 12 39 49
The student counter incorrectly shows the number of unique student IDs that joined a classrooms. However, because students get a unique ID when they join a classroom, even when the student is the same, it means that the number currently shown doesn't reflect the number of actual unique students, as it includes students with the same username.
Solution: filter members array by username.

shaunanoordin:
Fixes #153 
- Fixes the Student counter on the Teacher Overview page; the counter now only counts unique Zooniverse IDs/logins.
- The redux state now also keeps track of one new variable: `classrooms.uniqueMemberLogins`, which is a String array.

@simoneduca and @eatyourgreens, this is ready for review. Simone, if you want to suggest a better variable name than `uniqueMemberLogins`, do let me know - I realise my naming convention tends to be wordier than yours.

### DIFF
--- a/src/presentational/ClassroomsOverview.jsx
+++ b/src/presentational/ClassroomsOverview.jsx
@@ -38,7 +38,7 @@ const ClassroomsOverview = (props, context) => (
               </div>
               <div className="col-xs-9 text-right">
                 {(context.classrooms.loading === false) ?
-                <h1>{context.classrooms.members.length}</h1>
+                <h1>{context.classrooms.uniqueMemberLogins.length}</h1>
                 : <Spinner/>}
               </div>
             </div>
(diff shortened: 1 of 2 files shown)


simoneduca:
I'd say `uniqueMembers` perhaps?

shaunanoordin:
Agreed: `uniqueMemberLogins` now renamed to `uniqueMembers`!

Also, as per the Slack discussion, we'll leave `members` alone for the time being.

simoneduca:
Thanks! LGTM :+1:

PR #171, opened 2016-04-07
Fix 144, 169, 170 - Login-secured Join Classrooms, Secure page fix, Navbar capitalisation compatibility

### LINKED ISSUE #144: Students/Explorers can access Join Classroom page when not logged in
This is primarily an issue because the Join Classroom button assumes the user is already logged in.

Will be fixed by a resolution to #75 and setting the Student/Explorer pages to Authentication-protected.

### LINKED ISSUE #169: Login can be bypassed by typing in URL!
I keep finding new holes to poke in the Authentication system.
- User not logged in, navigates to Teacher page: prompted to Login, OK.
- User logged in, navigates to Teacher page: page loads, OK.
- User logged in, refreshes page at Teacher page: page loads, OK
- User logged in, accesses /teachers/classrooms via typing in URL: page loads, OK
- User **not logged in**, accesses /teachers/classrooms via **typing in URL**: **page loads, ERROR**. _Should be prompted to Login!_

This is a problem if user bookmarks, e.g. https://learn.wildcamgorongosa.org/teachers/classrooms and then accesses it while not logged in.

Tangentially related to #144 .

### LINKED ISSUE #170: Navigation doesn't recognise URLs with StRanGe CaPitalISAtions
You can go to `https://learn.wildcamgorongosa.org/students/data`, but the top bar navigation won't highlight the 'Data' tab if the URL is `https://learn.wildcamgorongosa.org/STUDENTS/DATA`

shaunanoordin:
Fixes #144 - Join Classroom page is now login-secured
- **NOTE** we need to revisit the flow for this, because after a Student logs in, they won't be looking at the same "Join Classroom" page when they return from Panoptes. This has been an issue since the start, but this login prompt makes it that much more obvious.

Fixes #169 - Secure pages are now secure again
- If a non-logged user accesses a secured URL by typing in the URL directly, e.g. http://learn.wildcamgorongosa.org/teachers/data, they will now be properly prompted to log in.
- This was caused by a bug from #160 where `Layout.verifyLogin` was looking for `this.props.loginInitialised` instead of the correct `props.loginInitialised`, which mattered when `verifyLogin` was called by `componentWillReceiveProps`

Fixes #170 - NavBaR loVeS aLl CapitaliSAtioNS
- Navbar now recognises paths in all capitalisations, so /students/data/ works as fine as /STUDENTS/dAtA/

@eatyourgreens , this is ready for review.

### DIFF
--- a/src/containers/Students.jsx
+++ b/src/containers/Students.jsx
@@ -1,13 +1,17 @@
 import { Component } from 'react';
 
-import Layout from '../presentational/Layout.jsx'
-
+import Layout from '../presentational/Layout.jsx';
 
 export default class Students extends Component {
 
   render() {
+    let loginSecured = this.props.navItems.reduce((prev, item, index, arr) => {  //Is this particular sub-page login secured?
+      console.log(prev, item.loginSecured, this.props.location.pathname.toLowerCase().startsWith(item.to), item.to);
+      return prev || (item.loginSecured === true && this.props.location.pathname.toLowerCase().startsWith(item.to));
+    }, false);
+    
     return (
-      <Layout {...this.props} navItems={this.props.navItems}>
+      <Layout {...this.props} loginSecured={loginSecured} navItems={this.props.navItems}>
         {this.props.children}
       </Layout>
     );
@@ -20,6 +24,7 @@ Students.defaultProps = {
     {
       label: 'Classrooms',
       to: '/students/classrooms',
+      loginSecured: true,
     },
     {
       label: 'Data',
--- a/src/containers/Students.jsx
+++ b/src/containers/Students.jsx
@@ -6,7 +6,6 @@ export default class Students extends Component {
 
   render() {
     let loginSecured = this.props.navItems.reduce((prev, item, index, arr) => {  //Is this particular sub-page login secured?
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Looks good, but I'm not sure why `verifyLogin` is being passed props, rather than accessing the instance properties via `this`. Maybe something to revisit in future.

shaunanoordin:
Ah! You actually spotted the cause of the issue, @eatyourgreens: `verifyLogin` needs to be explicitly passed props because one of its calling functions, `componentWillReceiveProps`, lives in a strange no-man's land that exists at the cusp of the transition between an old Redux store state and a new Redux store state.

Explanation: When `Layout.componentWillReceiveProps(nextProps)` triggers for the first time (i.e. when the app checks if the user is logged in or not)...
- `this.props.loginInitialised` will be `false`, reflecting the Redux store state riiight before the app checks if the user is logged in, but
- `nextProps.loginInitialised` will be `true`, correctly reflecting that the state has been updated and the app has already checked if the user is logged in.

The bug was that, because verifyLogin kept checking **this.props**, it kept seeing the 'old' `loginInitialised` which was `false`, even though the app has checked the user login.

PR #180, opened 2016-04-12
Create separate Redux stores for Student.classrooms and Teacher.classrooms

### LINKED ISSUE #179: Teacher Classrooms carried over to Student Classrooms and Vice Versa
Issue: If you go to Teachers/Classrooms, you'll see a list of Classrooms _created_. If you then navigate to Students/Classrooms, you'll see _the same list_, instead of the intended list of Classrooms _joined_.
The reverse applies - if you go to Students/Classrooms first, then navigate to Teachers/Classrooms, then the list will show Classrooms _joined_ instead of Classrooms _created_.

Analysis:
- Essentially, there are two 'fetch Classrooms' function, one for Teachers, one for Students.
- Both will overwrite the same `reduxState.classrooms` variable.
- Once a fetch Classroom is called, the other won't be called again.

Solutions:
1. Long term: `reduxState.classrooms` needs to be split into `reduxState.teacher.classrooms` and `reduxState.student.classrooms`.
2. Quick fix: when user clicks on the Header Logo (the only way to navigate out of Student/Teacher modes), then set the URL to `window.location.host` instead of doing a React-Router transition. No worries if the user types in the URL directly - this will force the fetch data to reload anyhoo.

shaunanoordin:
**Intent:** Fix #179 

*Once this branch gets up and running we should have a separate redux store for Student classrooms and Teacher classrooms.
- Currently we have `reduxStore.classrooms` which is contested by 'fetch Teacher classroom' actions and 'fetch Student classrooms' actions. (i.e. listing which classrooms the user created, and which classrooms the user has joined.)
- This will then become `reduxStore.teacher.classrooms` and `reduxStore.student.classrooms`.


shaunanoordin:
This is not yet ready for review. As of [commit 4b825f9](https://github.com/zooniverse/wildcam-gorongosa-education/pull/180/commits/4b825f9ec95cc1867c5f90c3c03c16bca6c98b5b), we have managed to spin off a new set of `student` actions & reducers.

Next: turning the remainder of `classrooms` into `teacher.classrooms`.

**Seeking Advice**: Currently, there is a small coding issue with adding another layer of depth to the `reduxState` data store, from `reduxState.classrooms` to `reduxState.students.classrooms`. This is the issue:
- With only 1 layer of depth, the reducers can use...

```
      return Object.assign({}, state, {
        loading: false,
      });
```

...to update a single variable (loading) without changing the rest of the variables.

However, once we go one level deeper, we'll need to re-specify the _whole object_. 

```
      return Object.assign({}, state, {
        classrooms: {
          loading: false,
          data: action.data || [],
          error: action.error,
        }
      });
```

If there's a more efficient way of doing this, I'm all `ears.addEventListener('advise')`

shaunanoordin:
OK, we now have things working for teachers.

shaunanoordin:
Thanks for your patience, everyone. OK, this is a relatively large change to _ensure everything works the same_:
- `actions/classrooms.jsx` split into `actions/teacher.jsx` and `actions/student.jsx`
- `reducers/classrooms.jsx` split into `reducers/teacher.jsx` and `reducers/student.jsx`
- Redux store `redux.classrooms` now split into `redux.teachers.classrooms` and `redux.students.classrooms`
- All components referring to  `redux.classrooms` now refer to the specific teacher/student classroom (check the various `mapStateToProps()` of container components)

For testing, I've confirmed that
- Teachers can view their list of classes and access single classes (By navigation or direct URL in most cases, although note that bug #181 is still in effect)
- Students can view their list of classes and access single classes
- Teachers can create new classes
- Students can join classes

@eatyourgreens and @simoneduca, this is ready for comments & review.

eatyourgreens:
Looks good to me but I don't fully understand multiple stores vs. single stores in Redux, so I'm going to read this and think about it: http://stackoverflow.com/questions/33619775/redux-multiple-stores-why-not

eatyourgreens:
:+1: looks  good to me

PR #182, opened 2016-04-15
Allow access to specific classrooms by URL

shaunanoordin:
Fixes #181 
Requires #180  
- (Logged in) users can now access specific Teacher and Student classrooms by typing the URL. e.g. https://learn.wildcamgorongosa.org/teachers/classrooms/38
- This also fixes the bug where the app crashes when you reload the page when viewing a specific Teacher/Student classroom.
- Quite simply, for the 'view specific classroom' components (`containers/Classroom.jsx` and `containers/StudentClassroom.jsx`) we simply added a check to see if the classroom data has loaded or not. If no, we present the 'loading spinner' while the data loads. (Previously, the app would just crash, expecting fully loaded data.)

This PR is ready for review, but PLEASE DO NOT MERGE IT YET. This PR is based off PR #180 (i.e. branch `fix-179`), so that needs to be merged first . Once that's done, please let me know and I'll rebase `fix-181` onto the new `master`.

(Alternatively, I can imagine merging this `fix-181` branch into the `fix-179` branch, then merging that into `master`, but that requires more Git voodoo than I'm comfortable with.)

PR #184, opened 2016-04-20
Allow access to specific classrooms by URL

### LINKED ISSUE #181: Accessing Classrooms by ID/URL doesn't work
Issue: Directly accessing the URL of a specific classroom, e.g. https://learn.wildcamgorongosa.org/teachers/classrooms/38 (Teachers) or https://learn.wildcamgorongosa.org/students/classrooms/10 (Students), will cause a perpetual 'page is loading' problem. 
Analysis:
- Console Error: `Uncaught (in promise) TypeError: Cannot read property 'students' of undefined`
- Error (for Teachers) can be traced to this bit in `src/presentational/Classroom.jsx`:

```
  render() {
    const { attributes} = this.props.data;
    const allMembers = this.props.members;
    const classroomMembers = this.props.data.relationships.students.data;  //This bit here
    return (
```

eatyourgreens:
Fixes #181 
Suprecedes #182
- (Logged in) users can now access specific Teacher and Student classrooms by typing the URL. e.g. https://learn.wildcamgorongosa.org/teachers/classrooms/38
- This also fixes the bug where the app crashes when you reload the page when viewing a specific Teacher/Student classroom.
- Quite simply, for the 'view specific classroom' components (containers/Classroom.jsx and containers/StudentClassroom.jsx) we simply added a check to see if the classroom data has loaded or not. If no, we present the 'loading spinner' while the data loads. (Previously, the app would just crash, expecting fully loaded data.)

### DIFF
--- a/src/containers/Classroom.jsx
+++ b/src/containers/Classroom.jsx
@@ -2,8 +2,7 @@ import React, { Component, PropTypes } from 'react';
 import { connect } from 'react-redux';
 
 import { default as ClassroomPresentational } from '../presentational/Classroom.jsx';
-
-
+import Spinner from '../presentational/Spinner.jsx';
 
 export default class Classroom extends Component {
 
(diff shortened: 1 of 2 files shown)


shaunanoordin:
The fix works as advertised - LGTM!

aliburchard:
nice!

PR #188, opened 2016-04-22
Allow editing of classroom details

### LINKED ISSUE #173: Add ability to update classroom data
Add edit button next to Classroom Name.
Use same form for creating a classroom, but pre-populated with existing data.

rogerhutchings:
Refactors the existing classroom creation form and allows it to be used for editing the classroom details. Tidies up some existing code as well. You might want to tweak the location of the edit classroom link, but doesn't really matter until we get a design done.

Fixes #173 

TODO:
- [x] Add support for editing existing classrooms to the educationAPI (@marten)
- [x] Finish the `editClassroom` action creator when that's done
- [x] Add a redirect after saving is finished


src/presentational/NewClassroomFormElement.jsx
```diff
@@ -0,0 +1,22 @@
+import React, { PropTypes } from 'react';
+import { Link } from 'react-router';
+
+const NewClassroomFormElement = props => {
+  const { label, ...other } = props;
+  other.name = other.name || label.toLowerCase();
+
+  return (
+    <div className="form-group">
+      <label>{ label }</label>
```
eatyourgreens: Label here isn't associated with the input element. It should probably wrap around the input.

rogerhutchings: Knackers the styling :( And IDs don't work, as they're getting broken by react-a11y. Presumably that's disabled in production?

eatyourgreens: Styling can be fixed by putting a span around the label text. See be5ffb20726d0c0a6c874a9a11f118b77906f2eb

rogerhutchings: Sound, I'll merge that in

simoneduca: Also, @rogerhutchings: yes, a11y is disabled in production.

simoneduca:

shaunanoordin:
This is weird - I tried testing the branch after an `npm update`/`npm install` but the app wouldn't run; so, I tried to get a baseline by running the `master` branch after deleting and reinstalling _all_ `node_modules`.

I'm rewarded by an error message: `Uncaught TypeError: Cannot read property 'init' of undefined`

Specifically, with `import { oauth } from 'panoptes-client';` in `Index.jsx`, the `oauth` object is undefined. Not sure what's happening here; investigations are ongoing.

shaunanoordin:
Weird... OK, so in `Index.jsx`, I had to change...
`import { oauth } from 'panoptes-client';`
to...
`import oauth from 'panoptes-client/lib/oauth';`

This was changed at the `master` branch, not the `edit-form` branch, so it's technically separate from this PR... but it has a chaining effect. At least now I can test this branch properly.

simoneduca:
@shaunanoordin the reason why that was happening is that the panoptes-client has gone through several changes recently, including the one you spotted regarding the path in the `import` statements. It affects all client modules. Out of curiosity, why did you do an `npm install`, when you checked out this branch?

shaunanoordin:
Thanks, @simoneduca! To recap our conversation:
1. I'll switch PR #210 to merge onto this branch instead of `master`
2. Q: why did I delete my `/node_modules`? A: Every once in a while, I 'reset' my project to ensure the repo runs correctly and doesn't rely on implicit dependencies that only work on my localhost, e.g. how my localhost had an older version of Panoptes Client than what the project actually demanded.

shaunanoordin:
@simoneduca, I've found a few issues, but only one is blocking.

Testing done on branch `fix-panoptes-client-mk2` (see #211), using Chrome50 + OSX.

Issue 1: After using Edit Classroom and clicking Submit, the list of Classrooms does not update to reflect the changes.
Analysis:
- Refreshing the page, however, shows that the Classroom List is properly updated.
- This indicates that the API was called correctly, but the return doesn't properly update the current `state.teachers` data

Status: **blocking**

Issue 2: With certain classes, trying to Edit Classroom and leaving certain fields blank will cause a console error. 
Analysis:
- Error message is 'NewClassroomForm.jsx?26ee:43 Uncaught TypeError: Cannot read property 'trim' of null'
- This is because with a few classes - created during the primordial days of the Education API - an empty 'description' or 'subject' wouldn't be a `""` empty string, but a `null`. As a result, calling `null.trim()` in `NewClassroomForm.handleSubmit()`'s `newClassroom[key] = this.state[key].trim();` would throw an error.

Status: non-blocking; will-not-fix is OK (Newly created classrooms will NOT have this issue; they will accept empty descriptions/subjects/etc just fine.)
screen shot 2016-05-20 at 14 07 44

Issue 3: Accessing the Edit Form directly, e.g. http://localhost:3000/teachers/classrooms/172/edit , will result in a console error.
Analysis:
- Error message is 'warning.js?0260:45 Warning: Failed propType: Required prop `classroom` was not specified in `EditClassroom`. Check the render method of `Connect(EditClassroom)`.'
- This is only a warning; the app works fine despite it.

Status: non-blocking, low-priority

I'm continuing to review the code, and also looking at how Issue #1 can be solved.

simoneduca:
@shaunanoordin I've merged your fix. Should be ready to go now.

shaunanoordin:
Thanks @simoneduca, this branch is now good to fly!
- Teachers can now edit Classrooms details
- Tested on Chrome50+OSX

Status: **Verified**
Actions: **merging**

PR #191, opened 2016-05-03
Add scrollbar to sidebar.

simoneduca:
Saves users from scrolling up and down the screen in order to see a classroom details, when there are many classrooms in the sidebar.


shaunanoordin:
- Scrollbar functions as advertised.
- I noticed the sidebar gets cramped visually when the window height is too small (approx <= `.admin-sidebar`'s `.max-height` of 500px), which is similar to the "component doesn't snugly fit parent container" problem that I encountered with MapExplorer; I can see possible improvements, but this PR works fine as is.

Tested on Chrome50/FF45/Safari9 + OSX

Status: **LGTM**
Actions: **Merging**

PR #194, opened 2016-05-04
Component.PropTypes renamed to Component.propTypes

shaunanoordin:
> Slack 27 Apr 2016 # wildcam-education
> Roger Hutchings 5:03 PM @ channel - working on some code for this atm. i’m seeing a lot of ComponentName.PropTypes - it should be ComponentName.propTypes (note case) otherwise it does nothing

As @rogerhutchings pointed out, some components were using the incorrect capitalisation for the `.propTypes` property; this PR fixes that.

### DIFF
--- a/src/presentational/CheckboxGroup.jsx
+++ b/src/presentational/CheckboxGroup.jsx
@@ -53,6 +53,6 @@ CheckboxGroup.defaultProps = {
   value: null,
 }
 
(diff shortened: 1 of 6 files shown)


simoneduca:
Ah, good catch! Easy to get confused though: https://facebook.github.io/react/docs/reusable-components.html#es6-classes

PR #195, opened 2016-05-05
Refactor MapExplorer to be Redux-friendly

### LINKED ISSUE #104: (Optimisation) Map Explorer to be refactored
As per @simoneduca's comments in https://github.com/zooniverse/wildcam-gorongosa-education/pull/102#issuecomment-199318498, Map Explorer and its subcomponents need to be refactored so it's/they're more streamlined with the rest of the app's Redux-centric pattern.
- States should be replaced with Redux stores
- DialogScreen component may be replaced with (react-modal)[https://github.com/reactjs/react-modal].
- Planned for the week of Mon 28 Mar to Fri 1 Apr 2016.

shaunanoordin:
Closes #104 

The Map Explorer was originally built on (a very convoluted implementation of) the basic React patterns. As the project has progressed however, we've embraced the new Redux-React patterns which makes storing data (and communicating data between components) much more easier.

This refactor attempts to:
- Make the MapExplorer code easier to digest by anybody who isn't the madman who wrote it in the first place (Yes, @shaunanoordin)
- Make the MapExplorer component(s) more in line with the React-Redux patterns used by other components in the project

I'm still in the process of making the code lighter & easier, so shout if you have any suggestions.


src/reducers/map.js
```diff
@@ -1,19 +1,33 @@
...
     case types.ADD_MAP_SELECTOR:
       let newSelectors = state.selectors.slice();
-      newSelectors.push(Math.floor(Math.random() * 100));
-      return {
-        selectors: newSelectors
-      };
+      newSelectors.push(new SelectorData());
+      return Object.assign({}, state, {
+        selectors: newSelectors,
+      });
```
simoneduca, line 13: @shaunanoordin You could use object spread operator here: 
`return {...state, selectors:  newSelectors,};`

simoneduca:
@shaunanoordin LGTM, but I've got a suggestion. In ES6, `var` is the least informative way of declaring a variable, and `let` or `const` should be preferred when possible. Here: https://github.com/zooniverse/wildcam-gorongosa-education/blob/mapexplorer-redux-refactor/src/containers/MapExplorer.jsx#L141, as far I can see, both `legend` and `div` are not reassigned, so I think they should be `const` instead of `var`.

shaunanoordin:
Thanks for the helpful feedback, @simoneduca ! I've gone through the Map Explorer code and improved the use of `const` and `let` across a few components. This PR is now ready for review!

simoneduca:
👍

PR #198, opened 2016-05-12
Student Classrooms Overview only count unique Classifications

shaunanoordin:
- What the title says

So here's the weird thing: what kind of numbers do we want to show on the Student Overview? What counts as my "total number of Classifications"?

Say I've joined 3 classes, then at WildCam Gorongosa, I make 5 separate Classifications. Technically I've only made 5 unique Classifications... but I did it (simultaneously) for 3 different classes. Does that mean I've made 15 total Classifications??

(Remember, at the moment, every class is for the same WildCam Gorongosa project, not different projects, where summing the Classifications across classes would make a wee bit more sense.)

This PR changes the behaviour of the Overview to ignores 'multiple' Classifications across every classroom and only counts unique Classifications - i.e. go through all the Classrooms the Student is a part of then pluck the biggest number. That sounds about right.

Ready for review.

### DIFF
--- a/src/presentational/StudentOverview.jsx
+++ b/src/presentational/StudentOverview.jsx
@@ -3,12 +3,19 @@ import Spinner from 'Spinner.jsx'
 
 const StudentOverview = (props, context) => {
 
-  let myClassifications = false;
   const {classrooms, user} = context;
+  
+  let myClassifications = false;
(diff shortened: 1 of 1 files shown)


simoneduca:
In light of today's conversation with @shaunanoordin, I'm going to, at least temporarily, close this. The reason is that we haven't decided yet how to properly show student classifications per classroom.

PR #199, opened 2016-05-13
Delete students

simoneduca:
This is waiting on the education-api to implementing the delete feature.
Note: this branch contains also the same code as `delete-classroom` because I originally branched off it.


simoneduca:
@eatyourgreens @shaunanoordin this is ready for review. Please note that you might experience unexpected behaviour while testing, i.e. you might not be able to delete certain classrooms. That is due to this Panoptes bug zooniverse/Panoptes#1811. I think we shouldn't consider that a blocker for merging this. Thoughts?

simoneduca:
Also, apologies for the multiple "resolve conflicts with master, etc" commits. Yesterday Github was playing up and I couldn't see whether commits were being pushed up or not.

src/presentational/StudentClassroom.jsx
```diff
@@ -11,7 +11,8 @@ export default class StudentClassroom extends Component {
 
   renderStudentList(allMembers, classroomMembers) {
     const list = (allMembers.length > 0) ? allMembers : [];
-    const studentIds = classroomMembers.map((student) => { return student.id; } )
+    const studentIds = classroomMembers.map(student => student.id)
+    console.log('studentIds', studentIds)
```
shaunanoordin, line 15: Is this `console.log` still necessary?

simoneduca, line 15: Nope, thanks.

shaunanoordin:
@simoneduca , I'm not sure if the Delete Students button does anything - I checked out `actions/teachers.js` and I don't see the Delete Students action actually talking to the API.

```
export function deleteStudent(classroomId, studentId) {
   return dispatch => {
     dispatch({
       type: types.DELETE_STUDENT,
       classroomId,
       studentId,
     });
     browserHistory.push('/teachers/classrooms/' + classroomId);
   }
 }
```

When I run the app and use the Delete Student function, the student is correctly removed from the classroom. But if I reload, the student is still there - it's as if the state is being modified, but not the database. I'm not sure if I'm missing something.

PS: You mentioned that _"you might experience unexpected behaviour while testing, i.e. you might not be able to delete certain classrooms"_ ; Noted, I'll only be testing deleting students in this PR, if that's a problem we'll investigate it as a separate issue.

simoneduca:
@shaunanoordin That's likely caused by that Panoptes bug. You should be able to confirm that by deleting that student's classroom. You should get a `500`.

simoneduca:
Thanks @shaunanoordin. Should be ok now.

shaunanoordin:
- Teachers can now Delete Students
- Students who are deleted from a Classroom will see that Classroom gone from their list
- Code looks clean
- No `500` Panoptes errors encountered.

Status: **LGTM**
Actions: **Merging**

PR #212, opened 2016-05-20
Teacher's Classroom List now reflects most recent Edited Classroom

shaunanoordin:
Addresses an issue brought up in #188 

Issue 1: After using Edit Classroom and clicking Submit, the list of Classrooms does not update to reflect the changes.
Analysis:
- Refreshing the page, however, shows that the Classroom List is properly updated.
- This indicates that the API was called correctly, but the return doesn't properly update the current `state.teachers` data

Solution: `EDIT_CLASSROOM_SUCCESS` action & reducers tweaked so that when the event triggers, the state's list of classroom is updated with the most recently (successful) edits.

@simoneduca , this is good to go. This PR can be reviewed & merged separately from #211 (i.e. 211 is _not_ a requirement for this to be merged into `edit-form`)

### DIFF
--- a/src/actions/teacher.js
+++ b/src/actions/teacher.js
@@ -59,6 +59,8 @@ export function editClassroom(fields, classroomId) {
     .then(response => {
       dispatch({
         type: types.EDIT_CLASSROOM_SUCCESS,
+        fields: fields,
+        classroomId: classroomId,
       });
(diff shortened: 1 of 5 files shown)


simoneduca:
LGTM, thank you!

PR #209, opened 2016-05-18
MapExplorer - May 2016 improvements [Rebased][READY FOR REVIEW]

shaunanoordin:
Let's take another stab at the MapExplorer
- We want a UI that's cleaner and easier to use than the existing version
- We want to leverage on some of CartoDB's Deep Insights functionality
- We want to remove all the unused stuff that's been hanging around since the initial alpha. (Plus code refactor.)
- Maybe... we want to reorganise the component names and file-folder structure, for orderliness' sake?

PR #219, opened 2016-05-27
IE10 + Edge Fixes: View Camera popup correctly size & popup dialogs close properly

### LINKED ISSUE #196: Map Explorer's View Camera is TINY on MS Edge
Issue: Subject photos in the Map Explorer's View Camera module look very small on Microsoft's Edge.

Analysis: This is what the View Camera module is supposed to look like (on FF)
screen shot 2016-05-06 at 17 41 25

This is what the module looks like on Edge
screen shot 2016-05-06 at 17 30 52

Notice that, in Edge, the size of the View Camera container box is tied to the size of the "Showing selected photos from..." text, and the photos then stretch to fit the size of the container box. So line of inquiry: what's controlling the size of the container box?

shaunanoordin:
Fixes #196
Fixes issue reported by @eatyourgreens on Slack:

> "Hmmm, if I open the camera popup on the map in IE10, I can’t close it again (or click on the map behind the overlay)"

Issue 1: The View Camera popup looks absolutely tiny in Edge and IE10
Analysis: In Edge/IE10, the container element (dialog-box ul) does not expand to fit its content (images)
Solution: Set a specific width for the container element

Issue 2: Attempting to close a View Camera or Download CSV popup dialog causes the app to crash

Analysis:
- The `DialogScreen-ViewCamera.closeMe()` function (and similarly for DialogScreen-Download) isn't properly bound to the `this` context.
- This is because the `closeMe.bind(this)` call is done in the parent class (DialogScreen) constructor, but the constructor isn't being called.
- This is because Babel doesn't properly transpile the `super()` call into a way that works on IE10 and below: https://phabricator.babeljs.io/T3041

Solution: Make the `super()`call redundant by making sure the parent's (DialogScreen's) constructor is also called in the childrens' (DialogScreen-ViewCamera, DialogScreen-Download) constructors.

NOTE: Screw DialogScreen-Species, it's being made redundant in PR #209 

Tested on IE10/Edge + Win7 (VirtualBox), Chrome50 + OSX

Status: Ready for review, @simoneduca and @eatyourgreens , but mind you, I'm doing some extra tests on IE10 and Edge to see what else if effing up.

### DIFF
--- a/src/presentational/DialogScreen-Download.jsx
+++ b/src/presentational/DialogScreen-Download.jsx
@@ -5,6 +5,7 @@ import {saveAs} from 'browser-filesaver';
 export default class DialogScreen_DownloadCSV extends DialogScreen {
   constructor(props) {
     super(props);
+    this.closeMe = this.closeMe.bind(this);  //Babel doesn't transpile super() properly in IE10, so we need to explicitly declare this.
     this.downloadCsv = this.downloadCsv.bind(this);
     this.blobbifyCsvData = this.blobbifyCsvData.bind(this);
     this.generateFilename = this.generateFilename.bind(this);
--- a/src/presentational/DialogScreen-ViewCamera.jsx
+++ b/src/presentational/DialogScreen-ViewCamera.jsx
@@ -4,6 +4,7 @@ import DialogScreen from './DialogScreen.jsx';
 export default class DialogScreen_ViewCamera extends DialogScreen {
   constructor(props) {
     super(props);
+    this.closeMe = this.closeMe.bind(this);  //Babel doesn't transpile super() properly in IE10, so we need to explicitly declare this.
   }
 
   render() {
--- a/src/styles/components/dialog-screen.styl
+++ b/src/styles/components/dialog-screen.styl
@@ -46,6 +46,7 @@
       list-style: none
       padding: 0
       margin: 0.5em 0 0 0
+      width: 80vw
       max-width: 1000px;
       
       li


src/presentational/DialogScreen-Download.jsx
```diff
@@ -5,6 +5,7 @@ import {saveAs} from 'browser-filesaver';
 export default class DialogScreen_DownloadCSV extends DialogScreen {
   constructor(props) {
     super(props);
```
shaunanoordin, line 7: I'm not entirely sure if I should remove this super() call, because on FF & Chrome it means that `this.closeMe = this.closeMe.bind(this)` will trigger twice. I mean, it _works_, but it's such a hack just to get things working for IE10.

src/presentational/DialogScreen-ViewCamera.jsx
```diff
@@ -4,6 +4,7 @@ import DialogScreen from './DialogScreen.jsx';
 export default class DialogScreen_ViewCamera extends DialogScreen {
   constructor(props) {
     super(props);
```
shaunanoordin, line 6: What I said above.

src/presentational/DialogScreen-Download.jsx
simoneduca, line 7: ~~`super()` must be used before the `this` keyword can be used, so I think it should stay in.~~
Sorry, I got that wrong. I don't think the wording on [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super) is very clear, but the bottom line is that you can't use `super()` after `this`, but, of course, you can use `this` without `super()`.

src/presentational/DialogScreen-ViewCamera.jsx
simoneduca, line 6: Same here.

simoneduca:
LGTM.

PR #222, opened 2016-06-07
Added an alert() for when Joining a Classroom fails (or succeeds)

### LINKED ISSUE #221: Students joining deleted Classrooms should receive an error message
Issue: A Student attempting to join a deleted/non-existent Classroom will fail, as expected, but the Student doesn't receive a message telling her/him why, which isn't user-friendly.
Expected Solution: Prompt an error message, obviously.

This is number 5 on my Delete-Classroom-testing notes. 
1. Teacher can create Classroom - OK
2. Student can join Classroom - OK
3. Teacher ​cannot​ join own Classroom - silent error
4. Teacher can delete Classroom - OK
5. Student cannot join deleted Classoom - OK, obviously, but silent error

shaunanoordin:
Fixes #221 
- When a Student attempts to join a deleted/non-existent Classroom, the attempt will fail (as expected), then display an error message, and then redirect normally to the Student Classroom Overview page.


shaunanoordin:
Derp, I was also supposed to find a more elegant solution to using `alert()`. As @simoneduca suggested, we may need to transform the 404 error page into a more general error page that we can redirect users to.

simoneduca:
@shaunanoordin it would be good to have a consistent way of displaying an error. If you're happy to take care of it, I won't merge this; otherwise this will do for now.

shaunanoordin:
A proper error page may be the best for the long term, and this isn't an time-urgent bugfix - I'll expand this PR, so please wait a bit before merging. Thanks, Simone!

shaunanoordin:
@simoneduca, I've rebased this solution and made updates so...
- A Student will receive a message when successfully joining a Classroom
- A Student will receive an error message when she/he cannot join a Classroom.

simoneduca:
LGTM!

PR #223, opened 2016-06-08
Update webpack, stylus deps

rogerhutchings:
This updates webpack to 1.13, which suppresses all the deprecated warnings. It also required a stylus-loader bump, which in turn needed an explicit stylus install.


simoneduca:
Nice one @rogerhutchings, this clears my console from all sort of rubbish :)

PR #224, opened 2016-06-13
Add Zooniverse Favicon

shaunanoordin:
We have a favicon!

zooniverse-icon-web-black

Two things are required for this to work:
1. The `loader` in `webpack.config.js` needs to use...

```
{
  test: /\.ico$/,
  loader: 'file?name=[name].[ext]',
},
```

...to load the `.ico` file without compressing it (which `file-loader` would do) and renaming it.
1. The entry file, `Index.jsx` needs to have some sort of reference to the `.ico` file so Webpack knows that the file is needed by the app. `import favicon from './images/favicon.ico';` works for this purpose.


simoneduca:
LGTM, thanks Shaun.

PR #225, opened 2016-06-27
Assignment form

simoneduca:
This basically me playing around with the concept of assignment.
It includes a workable assignment form, but ultimately assignments will have their separate store (actions and reducers). 
Since the idea of assignments has also be honed further thanks to chats with others in the team, I might close this PR and reopen a new cleaner one in the next day or so.


simoneduca:
I don't know what is going on, I keep having conflicts with master and `assignment-form` branch.

simoneduca:
@eatyourgreens @rogerhutchings @shaunanoordin if anybody feels brave this is ready. It's inoffensive stuff really. Note: the edu-api has a bug now, so when you create an assignment you get a `500` atm, but don't worry the call is correct.

src/actions/assignment.js
```diff
@@ -0,0 +1,67 @@
...
+            subjects: {
+              data: subjectData,
+            }
+          }
+        }
+      })
+    })
+    .then(response => response.json())
+    .then(json => {
+      console.log('json', json);
```
eatyourgreens: Leftover debugging line?

src/actions/assignment.js
```diff
@@ -0,0 +1,67 @@
...
+import config from '../constants/config';
+import * as types from '../constants/actionTypes';
+
+
+// Action creators
+const { root, assignments } = config.eduAPI;
+
+export function createAssignment(assignment, classroomId) {
+
+  const classroomData = Object.assign({}, {
```
rogerhutchings: why Object.assign and not a plain obj literal?

src/actions/assignment.js
```diff
@@ -0,0 +1,67 @@
...
+      id: classroomId,
+      type: 'classrooms',
+    })
+  const metadata = Object.assign({}, {
```
rogerhutchings: same here

src/actions/assignment.js
```diff
@@ -0,0 +1,67 @@
...
+    classifications_target: assignment.classifications_target,
+    description: assignment.description,
+    duedate: assignment.duedate,
+  });
+  const studentData = assignment.students.map(student_id => { return {id: student_id, type: 'student_user'} });
```
rogerhutchings: can be simplified to:

```
const studentData = assignment.students.map(student_id => ({ id: student_id, type: 'student_user' }));
```

src/actions/assignment.js
```diff
@@ -0,0 +1,67 @@
...
+  const subjectData = assignment.subjects.map(subject_id => { return {id: subject, type: 'subjects'} });
+
+  return dispatch => {
+    const createAction = { ...assignment, type: types.CREATE_ASSIGNMENT };
```
rogerhutchings, line 32: this isn't reused, so you might as well put it straight into `dispatch()`

src/actions/assignment.js
```diff
@@ -0,0 +1,67 @@
...
+      mode: 'cors',
+      headers: new Headers({
+        'Authorization': apiClient.headers.Authorization,
+        'Content-Type': 'application/json'
+      }),
+      body: JSON.stringify({
+        data: {
+          attributes: {
+            name: assignment.name,
+            metadata: metadata
```
rogerhutchings: can be just `metadata`

src/presentational/AssignmentForm.jsx
```diff
@@ -0,0 +1,144 @@
...
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.handleChange = this.handleChange.bind(this);
+    this.renderStudentList = this.renderStudentList.bind(this);
+    this.toggleStudent = this.toggleStudent.bind(this);
+    this.state = Object.assign({}, initialState);
+  }
+
+
+  toggleStudent(id) {
+    if (this.state.students.includes(id)) {
```
rogerhutchings: `const { students } = this.state`

then you can remove every instance of `this.state` in this function

src/presentational/AssignmentForm.jsx
```diff
@@ -0,0 +1,144 @@
...
+    } else {
+      this.state.students.push(id);
+      this.setState({
+        students: this.state.students,
+      })
+    }
+  }
+
+  handleChange(e) {
+    let nextState = {};
```
rogerhutchings: const here, assignment doesn't change

src/presentational/Dropdown.jsx
```diff
@@ -17,9 +21,10 @@ export default class Dropdown extends Component {
   render() {
     const options = this.props.options.map(option =>
         <option
-          key={option.value}
-          value={option.label}>
-            {option.label}
+          // the disjunctions allow for the case where options are classrooms (right-hand side)
```
rogerhutchings: what's going on here?

src/presentational/TeacherForm.jsx
```diff
@@ -1,12 +1,12 @@
...
-import Dropdown from '../presentational/Dropdown.jsx';
-import CheckboxGroup from '../presentational/CheckboxGroup.jsx';
-import RadioButtonGroup from '../presentational/RadioButtonGroup.jsx';
+import Dropdown from 'Dropdown.jsx';
+import CheckboxGroup from 'CheckboxGroup.jsx';
+import RadioButtonGroup from 'RadioButtonGroup.jsx';
 
 import { fetchUserDetails, upsertTeacherMetadata } from '../actions/users';
-import { age, boolean, courses, countries, resources, settings } from '../constants/util';
+import { age, boolean, courses,   countries, resources, settings } from '../constants/util';
```
rogerhutchings: weird spacing

shaunanoordin:
Woot, we're getting assignments! Nice, @simoneduca 

Here's some feedback and some queries.
1. Accessing the New Assignment feature is pretty straightforward for Educators, good.
2. [ui] The 'Date' field in the New Assignment form is _Month-Day_-Year? I'm in favour of YYYY-MM-DD, but I'm OK with this format if this suits our users' locale better. I'll make sure the date format across the rest of the website is consistent.
3. [error] Attempting to create a New Assignment with valid Name/Description/Date/NumOfClassifications and 1+ Students creates: `500 Error, https://education.staging.zooniverse.org/assignments/`
4. [error] However, when an attempt to create a similar Assignment with 0 Students creates a `422 Error (Unprocessable Entity), https://education.staging.zooniverse.org/assignments/`
5. Q: the wiring to the EduAPI is still a WIP?
6. [minor][ui] When clicking on a Student name to select ("assign") said Student, I'd recommend styling the pointer to `pointer: cursor`. (Or was that `cursor: pointer`? I can never remember.)
7. [accessibility] The list of selectable students in the New Assignment form is actually a list of `<td>`s, which might be improved if they were, e.g. `<button>`s... or perhaps a list of `<label><input checkbox>`s.
8. I have been unable to test Assignments from a Student perspective, unfortunately, due to points 3 and 4.

screen shot 2016-06-29 at 17 15 45
Note that hovering over the students zootester1, zootester2, zootester3 would not clearly indicate to the user that the names can be clicked.

src/actions/assignment.js
simoneduca: Because I'm an idiot.

src/actions/assignment.js
simoneduca, line 32: @rogerhutchings I think you used that pattern in `actions/teacher` for better readability (it's not reused there either). So I was trying to be consistent.

src/presentational/AssignmentForm.jsx
simoneduca: That's right. I used `let` because I was getting `nextState is read-only` on the next line. However I can't reproduce that now, bit confusing.

src/presentational/Dropdown.jsx
simoneduca: Sorry it's a bit confusing I guess. I originally written this component with the idea that the `options` were stored in separate file with `value/label` properties, i.e. `constants/utils.js`. However I wanted to be able to reuse the dropdown for classrooms as well, which don't have a `value/label` structure, hence the _ad hoc_ disjunction.

src/presentational/AssignmentForm.jsx
rogerhutchings: I imagine previously you were doing something like

```
const x = { foo };
x = { bar };
```

`const` means it should always point to the same reference, but the object it refers to is allowed to change. So here, you're changing a property on `nextState`, which is legit. _Replacing_ the entire state with a new state object is not

src/presentational/Dropdown.jsx
rogerhutchings: This probably isn't a great idea, as the end point is writing some mad conditional to accommodate everything that gets passed into it. Why not have a consistent API and modify the props passed in for classrooms to have a value/label structure instead?

simoneduca: Totes agree. It was something I did quickly, I'll have a go at that.

src/presentational/AssignmentForm.jsx
```diff
@@ -0,0 +1,144 @@
...
+        <InputElement
+          label="Number of classifications"
+          name="classifications_target"
+          onChange={this.handleChange}
+          placeholder="Note: per student"
+          required="required"
+          value={this.state.classifications_target}
+        />
+        <div className="form-group">
+          <label>Students</label>
```
eatyourgreens: This label isn't associated with a form input element.

src/presentational/AssignmentForm.jsx
```diff
@@ -0,0 +1,144 @@
...
+            <tr>
+              <th>Name</th>
+            </tr>
+          </thead>
+          <tbody>
+            {students.map((student) =>
+            <tr
+              className={this.state.students.find(id => id === student.id) ? 'success' : ''}
+              key={student.id}
+              onClick={this.toggleStudent.bind(this, student.id)}
```
eatyourgreens: Looks like you're using a `tr` as a checkbox here, which isn't good practice.

src/reducers/assignment.js
```diff
@@ -0,0 +1,43 @@
...
+  assignments: {
+    loading: false,
+    data: [],
+    error: false,
+  }
+};
+
+
+export function assignment(state = initialState, action) {
+  let newState = Object.assign({}, state);
```
rogerhutchings: `newState` isn't actually being used

src/presentational/AssignmentForm.jsx
simoneduca: It's not a checkbox @eatyourgreens. Is still bad practice?

eatyourgreens: It looks like it can be clicked with a mouse, and has a selected/deselected state, but `tr` is not a focusable element, so it can't be toggled with the enter or space keys. `tr` doesn't have a `checked` attribute either, so it won't announce its selected state to the browser APIs used by VoiceOver and other assistive tech. You can write a bunch more JavaScript to add this behaviour, but you'd get it all for free if you used a checkbox to select each student.

simoneduca: Cool, thanks. If I use a checkbox, would you recommend to get rid of the `success` class? The cue that the student is selected would be the `checkbox` now.

simoneduca:
@eatyourgreens @rogerhutchings @shaunanoordin  I've included your comments, thanks a lot.

src/presentational/AssignmentForm.jsx
```diff
@@ -0,0 +1,149 @@
...
+            {students.map((student) =>
+            <tr
+              className={this.state.students.find(id => id === student.id) ? 'success' : ''}
+              key={student.id}
+            >
+              <td>
+                {student.attributes.zooniverse_display_name}
+              </td>
+              <td>
+                <input type="checkbox" onClick={this.toggleStudent.bind(this, student.id)} />
```
eatyourgreens: These checkboxes still need labels. Maybe something like `<label><input type="checkbox" value={student.id} />{student.attributes.zooniverse_display_name}</label>`. Then clicking on the student name will select/deselect the checkbox.

simoneduca: @eatyourgreens pushed your suggested changes, would mind having a look?

eatyourgreens: @simoneduca cheers! The checkbox can be toggled by clicking on the student name now, but the `toggleStudent` handler isn't called. Also, it's possible to get the form into a state where the checkbox isn't ticked but the student is added to the classroom, and vice versa.

I'd suggest listening for changes to the checkbox, and maybe using React's implicit binding:

```
<label><input type=checkbox value={student.id} onChange={this.toggleStudent} />…</label>
…
toggleStudent(e) {
var student_id = this.value;
var selected = this.checked;
… logic for selecting/deselecting students goes here…
}
```

eatyourgreens: I'm an idiot. My snippet above should have used `e.target.value` and `e.target.checked`.

simoneduca: @eatyourgreens do you mean for `student_id` and `selected` respectively?
I think [React auto-binding is not available with ES6 classes ](https://facebook.github.io/react/docs/reusable-components.html#no-autobinding).

simoneduca:
@eatyourgreens hopefully the last commit should do it. Mind having a look?

simoneduca:
@eatyourgreens better?

PR #234, opened 2016-07-19
Move students to new module

rogerhutchings:
First of three PRs, next is teachers, then common. I've also tried to standardise things here and there


rogerhutchings:
Superseded by #235 and #236

Issue #229, opened 2016-07-12
Map Explorer - Download (and Save Subjects) don't update Map Selection before executing


shaunanoordin:
In Map Explorer, functions like `MapControls.prepareCsv()` use `this.updateMe(null);`

e.g.

```
prepareCsv(e) {
    //First things first: make sure the user sees what she/he is going to download.
    this.updateMe(null);
```

Issue: `this.updateMe(null)` doesn't change the selected Map Explorer data values during the `prepareCsv()` function itself.

So for example, if you load the page with the default Species Selected: All, then change it to Species Selected: Lions, then click Download, the CSV will be for Species Selected: All.

This is due to the way the lifecycle works, unfortunately. Suggestion: abstract the "get all the current selector data" from inside `updateMe()` into a separate `readLatestSelectorData()` function which returns a MapSelector data object (i.e. an object that contains all the current selection settings).

shaunanoordin:
Fixed in Commit 0e080c6b70f382c4d42f82b1bde3ba3bb758ea2a ... I can't remember which PR that went into, though. 230?

Issue #226, opened 2016-06-30
Specs: Creating New Assignments (and Linking Subject Sets to Them)


shaunanoordin:
Feature/User Story: Teacher wants to create a new Assignment. Obviously, the new Assignment needs Subjects attached to it.
### Method 1: Start at the Classroom->New Assignment Page

![wildcamlab-assignment-page1](https://cloud.githubusercontent.com/assets/13952701/16497741/6ec63696-3ef0-11e6-90ff-bdee7cf0a2ad.jpg)

If the user is a Teacher who just wants to create Assignments for her class, she should be able to have a straightforward flow from her Classrooms "control centre".
1. Teacher goes to an existing Classroom and chooses to create a New Assignment.
2. Teacher is taken to the New Assignment form.
3. After filling in all the fields, Teacher is prompted to 'create Subject Set' 
   - (Alternatively, Teacher can select from previously created Subject sets from a dropdown)
4. Clicking 'crete Subject set' takes the Teacher to the Map Explorer.
5. Teacher sets filters on the Map Explorer then clicks on the new 'create new Subject Set for Assignment' button **(new feature to be built)**
6. Teacher is then taken back to the previous New Assignment form, with the previous fields already filled, and the newly created Subject Set listed.
7. Teacher can now click the 'Create' button to confirm the creation of the Assignment.

Technical notes:
- When navigating out of the New Assignment form, the filled-in fields would need to be stored as a temporary object. (e.g. "Redux.store.teachers.TEMPNEWASSIGNMENT")
- When returning to the New Assignment form from the Map Explorer, the previously stored fields would be recalled to repopulate the form.
- Similarly, the newly created Subject Set would need to be stored somewhere, e.g. Redux.store.teachers.TEMPSUBJECTSET.
### Method 2: Start at the Map Explorer Page

![wildcamlab-assignment-page2](https://cloud.githubusercontent.com/assets/13952701/16497748/782242de-3ef0-11e6-8435-46ddc118d91a.jpg)

If the Teacher wants to explore the Map Explorer, she should be able to "save" the Subject Sets she finds interesting, then later use those sets in the Assignments she creates.
1. Teacher goes to the Map Explorer (Teacher->Data)
2. Teacher sets filters on the Map Explorer then clicks on the new 'save Subject Set' button **(new feature to be built)**
3. Teacher will then be taken to the Subject Sets admin page. **(new feature to be built)**
[...]

PR #240, opened 2016-07-20
Improve New Assignment creation process

shaunanoordin:
Intended Features: what it says on the tin.
- Refer to Method 1 in #226
- When a Teacher is Classrooms choosing to create a New Assignment,
  - She should be able to transition to the Map Explorer easily
  - then transition back to the New Assignment form (with all details stored)
  - then be properly notified that the Assignment is created.

Code Specifics:
- [x] `teachers/components/AssignmentForm.jsx` loads saved NewAssignment data on transition in.
- [x] `teachers/components/AssignmentForm.jsx` saves NewAssignment data on transition out to Map Explorer.
- [x] ...as well as the Classroom ID
- [x] Upon choosing "use Subjects for Assignment", Map Explorer redirects back to New Assignment form **if** it knows what Classroom ID it should push back to.
- [x] ...what to do when no Classroom ID specified? Consider. (UPDATE: we stick with the old "pls go 2 assignment pg" message)
- [x] `teachers/components/AssignmentForm.jsx` gives a proper success/error message upon the creation of an Assignment (UPDATE: we're using an `alert()`...)
- [x] ...and transitions users back to the respective Classroom page.

### DIFF
--- a/src/modules/teachers/actions/assignment.js
+++ b/src/modules/teachers/actions/assignment.js
@@ -60,11 +60,16 @@ export function createAssignment(assignment, classroomId) {
     })
     .then(response => response.json())
     .then(json => {
+      console.log('!'.repeat(40));
+      console.log(json);
+      console.log('!'.repeat(80));
+      console.log(json.data);
       dispatch({
         type: types.CREATE_ASSIGNMENT_SUCCESS,
         data: json.data,
       });
-      browserHistory.push( '/assignments/' + json.data.id);
+      browserHistory.push( `/teachers/classrooms/${json.data.attributes.classroom_id}`);
+      alert('Assignment created!');  //TODO: we need better messaging
     })
     .catch(response => console.log('RESPONSE-error: ', response));
   };
--- a/src/modules/teachers/actions/assignment.js
+++ b/src/modules/teachers/actions/assignment.js
@@ -60,10 +60,6 @@ export function createAssignment(assignment, classroomId) {
     })
     .then(response => response.json())
     .then(json => {
-      console.log('!'.repeat(40));
-      console.log(json);
-      console.log('!'.repeat(80));
-      console.log(json.data);
       dispatch({
         type: types.CREATE_ASSIGNMENT_SUCCESS,
         data: json.data,
--- a/src/modules/teachers/components/AssignmentForm.jsx
+++ b/src/modules/teachers/components/AssignmentForm.jsx
@@ -1,5 +1,5 @@
 import { Component, PropTypes } from 'react';
-import { Link } from 'react-router';
+import { Link, browserHistory } from 'react-router';
 
 import InputElement from './InputElement';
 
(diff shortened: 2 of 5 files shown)


src/modules/teachers/actions/assignment.js
```diff
@@ -64,7 +64,8 @@ export function createAssignment(assignment, classroomId) {
         type: types.CREATE_ASSIGNMENT_SUCCESS,
         data: json.data,
       });
-      browserHistory.push( '/assignments/' + json.data.id);
+      browserHistory.push( `/teachers/classrooms/${json.data.attributes.classroom_id}`);
+      alert('Assignment created!');  //TODO: we need better messaging
```
shaunanoordin, line 68: `alert(' This makes me sad. :( ');`

shaunanoordin:
@rogerhutchings , this is ready for review! I've tested this on FF+OSX but I'm still running the update through more cross-browser tests. I know Safari's Private mode, for example, loves to screw around with localStorage/sessionStorage.

shaunanoordin:
Added two fixes:
- Assignments are now filtered by Classroom, so e.g. Classroom Apple's Assignments won't appear in Classroom Cherry's list.
- Fix Issue #229 - in the Map Explorer, Download and Save Subject now properly take the value of the current (updated) filter controls before executing.

src/modules/teachers/components/AssignmentForm.jsx
```diff
@@ -63,21 +69,23 @@ class AssignmentForm extends Component {
...
           '4080',
-          '4075'
+          '4075',
         ];
-    }*/
+    }
     //--------
     
     //NOTE: According to Marten, it's perfectly OK to create a subject with no students or subjects.
     //--------
```
simoneduca, line 85: You mean create an _assignment_? @shaunanoordin

src/modules/teachers/components/AssignmentForm.jsx
```diff
@@ -63,21 +69,23 @@ class AssignmentForm extends Component {
...
     if (newAssignment.students.length > 0 && newAssignment.subjects.length) {
+      sessionStorage.setItem('savedNewAssignment', null);
+      sessionStorage.setItem('savedClassroomId', null);
```
simoneduca: The values of `savedSubjectsIDs` and `savedSubjectsLocations` should also be cleared.

src/reducers/assignment.js
```diff
@@ -13,7 +13,7 @@ export function assignment(state = initialState, action) {
     case types.CREATE_ASSIGNMENT:
       return { ...state,
         assignments: {
```
simoneduca, line 15: The shape of state is unnecessarily deep: `assignment`>`assignments`>`data` should just be `assignments`>`data` or similar. @shaunanoordin I realise that's actually my doing, but it'd be nice if we didn't end with awkward chains like `state.assignment.assignments`

src/modules/teachers/components/AssignmentForm.jsx
shaunanoordin, line 85: DERP. Gonna fix this.

src/reducers/assignment.js
shaunanoordin, line 15: Agreed; this looks ripe for a future code cleanup.

src/modules/teachers/components/AssignmentForm.jsx
shaunanoordin: I derped here too; never use `sessionStorage.setItem('whatever', null);` because `sessionStorage.getItem('whatever')` will return `"null"`, the string. Same with `undefiend`.

This should be `sessionStorage.removeItem('whatever')`.

simoneduca:
LGTM, thanks!

PR #250, opened 2016-07-27
View assignments

simoneduca:
Atm only the assignment name and students are shown. The code for viewing/editing is there but I'm waiting on a couple of education API fixes to be able to show all the data.


shaunanoordin:
- Code LGTM
- As a Teacher, I can access Classrooms -> Assignments and View specific Assignments.
  - As a Teacher, I can edit the Name, Description, Date and Number of Classifications of an Assignment.
  - Ability to change assigned Students and assigned Subjects have not been implemented yet, as expected.
  - FYI, View/Edit Assignments will NOT work when viewing old Assignments that were created before the Metadata EduApi fix.

Status: LGTM
Actions: Merging

PR #244, opened 2016-07-25
Map Explorer (Episode 1) - Add Time of Day, Dist to Water, Dist to Humans filters

shaunanoordin:
Features:
- Users of the Map Explorer can now filter results by...
  - Time of Day (Dawn, Day, Dusk, Night)
  - Distance to Water
  - Distance to Humans

This addresses _"Add filters for: time of day (day, night, dawn, dusk); distance to human structure, distance to water source. TOP-PRIORITY."_ in #242 

This is the basic version (no fancy UI), so [you get a textbox, you get a textbox, you get a textbox, _everybody_ gets a textbox.](https://youtu.be/8CAscBCdaQg?t=1m32s)

Ready for review!

Issue #242, opened 2016-07-25
Milestone - AUG 5


aliburchard:
@shaunanoordin @simoneduca @rogerhutchings @eatyourgreens 

I wanted to update the milestone brain dump to discrete priorities that are critical for launching the Lab publicly. Some of these are big, some are small, but I'd like to try and get these done by Aug 5, so we can test before HHMI announces the lab publicly on August 15.

THere's tons of exciting things to do afterwards, but we need to get these bits done first.

MAP EXPLORER
- [x] Add filters for: time of day (day, night, dawn, dusk); distance to human structure, distance to water source. **TOP-PRIORITY.**
- [x] see #245 Clean up CSV outputs: make column headers intuitive. Bridget will provide headers.
- [x] see #247 add layers for water, human stuff, etc.  (bridget resending layers)
- [x] Display counter of sites and images that you've selected (e.g. once you select "wildcat" you should see how many images meet those criteria and how many different sites they are at) 

ASSIGNMENTS
- [x] #257 View assignments (students)
- [x] Allow students to download their own data (falls within above issue, but really important)
- [x] #252 View assignments (teachers). 
- [x]  Delete assignments (teachers) 
- [x]  disable submit until finished & add indicator for processing 
- [x] Allow assignment descriptions to be longer. Expect more like a paragraph, that might include some guidance for students. (So teachers might essentially write instructions to their students.) 
- [x] remove pre-populated images during assignment creation. 
- [x] due date should be optional 
- [x] limit number to numeric 
- [x] **NEW** Add button for "download aggregated data" from assignment next to "download my data" from an assignment

OVERALL
- [x] Add tutorials
- [x]  #249 Create additional educational content pages. Bridget has sent content. These will be basic just like the educational resources page.

aliburchard:
I'm editing this to remove lower priority items (I'll move them to a wish list). 

Based on internal testing, I'm going to make another list with high priority items.

aliburchard:
okay, based on conversations just now, I'm taking "clone assignments" off the critical list.  We have one new priority item (downloading aggregated data) instead.

Ordered priorities:
1) disable submit button while thinking (let me know if this is harder than a couple hours, @simoneduca) 
2) add "download my aggregated data" button (for @simoneduca after submit button is disabled)
3) add tutorials (@shaunanoordin, after you submit your map explorer PR)
4) add bridget's extra content (whoever gets there first).

simoneduca:
[...]

PR #251, opened 2016-07-28
Map Explorer (Episode 3) - Gorongosa National Park boundaries

### LINKED ISSUE #247: Add layers to map
@shaunanoordin I think at the bare minimum we need park boundary to be always displayed. Ideally we would also have the ability to add vegetation type and switch between veg type vs. satellite or the other base layers that you have. I don't know how much harder it is to make something an alternative base layer vs. adding it as an optional layer on _top_ of the existing base layer. 

[Educator Portal Map.zip](https://github.com/zooniverse/wildcam-gorongosa-education/files/383709/Educator.Portal.Map.zip)

shaunanoordin:
Features:
- The geographical boundaries for Gorongosa National Park are now shown in Map Explorer.
- The boundaries can be toggled on/off using the Layers button (top right of the map)

screen shot 2016-07-28 at 17 46 02

Minor fixes:
- "Select for Assignments" button, on Map Explorer, no longer appears if you're a student. Because that would be dumb.

Closes #247

Addresses the minimum for 'add layers for water, human stuff, etc' on #242 with room for adding mooore layers.

Ready for review, but please either...
- merge #244 first, then #248 second, then review this finally,
- OR ignore #244 + #248, and only merge this branch as a "combined merge". (This branch is based on 244 + 248.)


aliburchard:
👏

simoneduca:
LGTM 👍

PR #253, opened 2016-08-01
Add multiple lines description input.

simoneduca:
Makes due date optional.
Limit classifications to number type.


shaunanoordin:
- Confirmed new restriction rules for Assignment: optional due dates, numbers-only Classifications Count
- Confirmed larger box for Description text.

Status: LGTM!
Actions: Merging

PR #255, opened 2016-08-01
Count selected subejcts for an assignment.

simoneduca:
NOTE: works only for newly created assignments (always showing 6 because we're on staging)

### DIFF
--- a/src/modules/teachers/actions/assignment.js
+++ b/src/modules/teachers/actions/assignment.js
@@ -18,6 +18,8 @@ export function createAssignment(assignment, classroomId) {
     classifications_target: assignment.classifications_target,
     description: assignment.description,
     duedate: assignment.duedate,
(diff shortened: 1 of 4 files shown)


simoneduca:
Actually number of subjects will not show after a fresh fetch. Derp.

simoneduca:
@shaunanoordin I fixed this now. All ready for you.

shaunanoordin:
- After an Assignment is created, a Teacher can access the assignment again via View/Edit and see the list of Map 'filters' used (i.e. Map Explorer Selectors data) as well as number of subjects selected.
  - This information appears to be stored in the `metadata` of `assignments` in the EduAPI.
  - As a result, this feature only applies to newly created Assignments; any Assignments created prior to this would not have the proper `metadata` information

Status: LGTM
Actions: Merging

PR #256, opened 2016-08-02
Map Explorer (Episode 4) - Improved GeoData Layers

shaunanoordin:
Features
- Gorongosa National Park geodata layer restyled so it's only a border.
- Vegetation geodata layer restyled to support different colours for each terrain type.
  - No new geodata layers added to avoid noisiness on Map Explorer... for now.
- Legend for Vegetation geodata layer added.
- The "Layers" submenu, on the upper right, is now permanently expanded to make it SUPER obvious to users that they can toggle layers.

screen shot 2016-08-02 at 15 30 56
Screen layout on a MacBook Pro (Retina, 13-inch, Early 2015). Also note use of colours; the mountain is pink-ish because I simply ran out of contrasting colours that work at such a low opacity.

Known Issue
- CartoDB Data layer appears below the GeoJSON overlay layers.
- This is because the CartoDB is technically a raster 'tile' layer that is grouped in Leaflet's `tilePane` (together with the base map layer), instead of the 'higher' (z-index-wise) `overlayPane` where the GeoJSON layers appear. So... rearranging them would be impossible unless I move the CartoDB Data layer out  of the `tilePane` group somehow.
- Workaround: GeoJSON overlay layers use visual tricks to make it less obvious that the CartoDB data points are beneath them. (Low opacity, etc)

This PR is ready for review. Improves on closed issue #247 and the 'add layers item' on milestone list #242

### DIFF
--- a/src/modules/common/containers/MapVisuals.jsx
+++ b/src/modules/common/containers/MapVisuals.jsx
@@ -92,8 +92,8 @@ class MapVisuals extends Component {
     
     //Create the CartoDB Data layer
     cartodb.createLayer(this.state.map, config.cartodb.vizUrl)
-      .addTo(this.state.map)
-      .on('done', (layer) => {
+      .done((layer) => {
+        layer.addTo(this.state.map);
         this.state.cartodbLayer = layer;
         this.state.cartodbLayer.setInteraction(true);
         layer.on('error', (err) => {
@@ -102,14 +102,16 @@ class MapVisuals extends Component {
 
         //Add the controls for the layers
         L.control.layers(baseLayersForControls, {
-          ...geomapLayers,
           'Data': layer,
+          ...geomapLayers
+        }, {
+          'collapsed': false,
         }).addTo(this.state.map);
 
         //updateDataVisualisation performs some cleanup
         this.updateDataVisualisation(this.props);
       })
-      .on('error', (err) => {
+      .error((err) => {
         console.error('ERROR (initMapExplorer(), cartodb.createLayer()):' + err);
       });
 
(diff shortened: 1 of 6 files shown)


src/modules/common/containers/MapVisuals.jsx
```diff
@@ -128,12 +128,25 @@ class MapVisuals extends Component {
     const legend = L.control({position: 'bottomright'});
     legend.onAdd = (map) => {
       const div = L.DomUtil.create('div', 'info legend');
-      div.innerHTML +=
+      div.innerHTML =
```
simoneduca: All this html should be put into a presentational component instead.

src/modules/common/containers/MapVisuals.jsx
```diff
@@ -128,12 +128,25 @@ class MapVisuals extends Component {
...
         '<div><svg height="10" width="10"><circle cx="5" cy="5" r="5" fill="#666" /></svg> : Camera with no images</div>' +
         '<div><svg height="10" width="10"><circle cx="5" cy="5" r="5" fill="#f93" /></svg> : Camera with images (click to view)</div>';
       return div;
     };
     legend.addTo(this.state.map);
+    
+    //Bonus: Add (vegetation) legends to map
+    const vegetationLegend = L.control({position: 'bottomright'});
+    vegetationLegend.onAdd = (map) => {
+      const div = L.DomUtil.create('div', 'info legend');
```
simoneduca: same as above

shaunanoordin:
@simoneduca I've addressed the feedback on the Map Legends

simoneduca:
LGTM 👍

PR #258, opened 2016-08-02
[EXPERIMENTAL] Map Explorer - Use CartoDB as GeoJSON

shaunanoordin:
Currently we have CartoDB as a Tile Layer in Leaflet. This has caused some... unique problems esp. in regards to the layering order with the new GeoJSON data, which appears on top of all tile layers.

GOOD NEWS! We can now retrieve the CartoDB data as GeoJSON data.

I've talked to Stuart of CartoDB about this (thanks dude!) and with only 200 POI markers at max (we have about that many cameras), we shouldn't run into any performance issues between raster (tile layer) vs vector (GeoJSON overlay)

### DIFF
--- a/src/map-data/gorongosa-geodata.json
+++ b/src/map-data/gorongosa-geodata.json
@@ -9,8 +9,8 @@
   },
   "options": {
     "style": {
-      "color": "#c93",
-      "opacity": 0.5,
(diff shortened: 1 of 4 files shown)


shaunanoordin:
Status: Superseded by #272 
Actions: Closing

PR #261, opened 2016-08-04
Show student progress to teachers.

simoneduca:


shaunanoordin:
- As a Teacher, I can view the Progress (Classifications made for specific Assignment) of each Student by going into Classroom > specific Classroom > Assignment > specific Assignment > View/Edit
- However, an **error** is detected when attempting to click Submit (i.e. editing an existing Classroom)
  - After clicking Submit, app 'crashes'
  - Error in console log is: `Uncaught TypeError: Cannot read property 'length' of undefined`
  - Error occurs during: `action: EDIT_ASSIGNMENT`
  - Possible point of error:

```
const getStudentClassifications = (student) => {
      if (this.props.student_data.length > 0) {  //this.props.student_data may be undefined
        const currentStudentData = this.props.student_data.find(item => item.attributes.student_user_id.toString() === student.id);
        return currentStudentData.attributes.classifications_count;
      } else {
        return 0
      }
    }
```

If this is the actual error source, then I'm not sure if `this.props.student_data` _should_ always have data, or if a simple `if (this.props.student_data && this.props.student_data.length > 0)` would do the trick.

shaunanoordin:
The last push fixed the error, but unfortunately I'm encountering a new one.

Issue: After successfully editing an Assignment as a Teacher, I can no longer view other Assignments.
Analysis:
- Steps:
  - As a Teacher, go into a Classroom, select an Assignment to View/Edit
  - Change the title/description/etc. Click Submit.
  - You will be taken back to the Classroom details as normal.
  - Go back to the Assignments tab and select an Assignment
  - Assignment will now say "Classroom Loading" with all details blank.
- Tested with `zootester1` using the `Durian` class.
- Note that reloading the page will set things back to normal, AND the changed title/description/etc actually applies properly. So the data seems to be going into EduAPI, but the callback afterwards... perhaps the Redux store for the Assignment got messed up?  
- Only occurs in `teachers-view-progress` branch; `master` branch looks OK. 
  Status: Investigating

shaunanoordin:
Fix confirmed!
- As a Teacher, I can edit an Assignment, then continue to view other Assignments as normal.

Status: LGTM!
Actions: Merging

PR #262, opened 2016-08-04
Map Explorer (Episode 5) - Summary text with results count

shaunanoordin:
Features
- The Map Explorer will now show a summary of the filter options selected and the number of results being shown.

screen shot 2016-08-03 at 16 33 55

Addresses 'Display counter of sites and images that you've selected' in #242


shaunanoordin:
Instead of merging this PR into `master`, we're merging it to `beta-test`

PR #266, opened 2016-08-09
Fixes #233

### LINKED ISSUE #233: Work out what's going wrong with newer versions of PJC
As per #231 - newer versions of PJC are producing auth errors. It _appears_ to be working when we pin the version to 2.2.1, but this is sub-optimal.
- [ ] Work out why newer versions of PJC break auth
- [ ] Add the caret back in front of `panoptes-client` in `package.json` to allow newer versions

simoneduca:
Uses `oauth` module for checking current user.
Reinstate the caret versioning for PJC.
Closes #233

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "json-loader": "^0.5.4",
     "mocha": "^2.3.0",
     "nib": "^1.1.0",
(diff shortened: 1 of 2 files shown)


simoneduca:
Thanks!

PR #272, opened 2016-08-11
Map Explorer ver3

shaunanoordin:
It's a whole new version of the Map Explorer, woo!

New Features:
- New UI!
  - Map filters (formerly "Map Selectors") are now in a collapsible module on the upper right of the map.
  - The **'View Filters'** button shows a list of filters, grouped by category.
  - Each filter group will spell out to the user if any filters options have been selected.
    screen shot 2016-08-11 at 21 01 15
    screen shot 2016-08-11 at 21 01 41
- The the visualisation of the results on the map (and the result count) are now **shown instantly**, as soon as you change any filter settings.
- The View Camera feature has more details about the camera itself.
  screen shot 2016-08-11 at 21 11 14
- The **Select for Assignments** feature has an improved interface.
  - Teachers are informed of how many results they have, and can choose a lesser amount to set for their Assignment. 
  - Compatibility with Assignments is retained (so I kept that messy String.join() method instead of using JSON.stringify(), oh well. Will optimiste interfaces/interlinks in another update.)
    screen shot 2016-08-11 at 21 01 56
    screen shot 2016-08-11 at 21 02 13

Old Features that are still there:
- The baseline map itself works mostly the same.
- Satellite/terrain layers
- Park-boundary and vegetation-type layers
- Download as CSV (now streamlined into a single button)

GONE:
- Excessive info dumps that used to exist surrounding the Download feature.

Known Issues:
- You'll see this warning message in the console: `Warning: _renderNewRootComponent(): Render methods should be a pure function of props and state; triggering nested component updates from render is not allowed. If necessary, trigger nested updates in componentDidUpdate. Check the render method of Script.`
  - This is caused by the incompatible(?) inclusion of the React components, `<MapLegendCameras />` and `<MapLegendVegetation />` into the Leaflet controls in `MapVisuals.jsx`
  - No idea how to fix this, asides from using `div.innerHtml` instead of Reach components. Help! 

Code Concerns:
- I created a new `modules/maps` folder for... yeah, for exactly what you think it does.
  - All the new code is in `modules/maps`, `helper/mapexplorer.js` and anything in `styles` that ends with '-ver3.styl'. This bits are standalone and shouldn't interefere with anybody else's code.
  - Changed code which maaay step on other PR's toes include `contants/mapexplorer.config.json`, `modules/teachers/index.js`, `modules/students/index.js`, `reducers/index.js`
- Hey @simoneduca, the map components have _finally_ been divided into smaller, more logical compos, and the lines of code have been made more manageable, with the largest compo having only 215 lines (and that's only because there's a lot of Leaflet stuff inside it). Aww yeah!
- I **did not delete the old MapExplorer** code. I'll need a separate PR to pull out that code.
- **Accessibility**:
  - I've ensured that the map is at least navigable via keyboard-only; the Filter options seem to work OK with the keyboard too. So, the components are keyboard-friendly...
  - My main accessibility concerns now, actually is making a unified keyboard-navigation experience across the React components. e.g. Tabbing from the Layers compo should go to the Filters compo (and not the Leaflet copyright links); ESC button should exit View Camera mode and back to Map navigation, etc.
  - I need to sit down and read up on how to use aria tags once the above is sorted; possible re-look into some of the more convoluted components to see if the HTML anatomy can be streamlined.

@simoneduca and @eatyourgreens, this is ready for review! @aliburchard, I'm also looking forward to your feedback too regarding the UI and UX. Wooo!


aliburchard:
👏 freaking amazing.

src/modules/students/index.jsx
```diff
@@ -17,6 +18,7 @@ const studentRoutes = store => (
         <Route path="assignments" component={StudentAssignmentsContainer} />
         <Route path="join" component={JoinClassroomContainer} />
         <Route path="data" component={MapExplorer} />
+        <Route path="data-ver3" component={MapExplorerVer3} />
```
simoneduca: Could we have a more meaningful name? Is `MapExlorer` going? @shaunanoordin

simoneduca:
LGTM, however we'll have to tackle this at some point https://github.com/zooniverse/wildcam-gorongosa-education/issues/275

PR #273, opened 2016-08-12
Prevent multiple clicking on assignment form submission.

simoneduca:


shaunanoordin:
- As a Teacher, I can go to Classroom > Assignment to create a New Assignment.
- After clicking the 'Submit' button, the button is disabled to indicate that the process of creating a New Assignment is in progress.
- At completion, an alert box will inform the user of the successful operation before redirecting the user to the Classroom page.
- No issues with the code

Status: LGTM
Actions: Merging

PR #277, opened 2016-08-12
Download aggregated data.

simoneduca:
No easy mode, i.e. all headers in the csv.


shaunanoordin:
- As a Student, I can go into Students > Assignments and either choose Download (download the classifications I made) or Aggregations (download the subject sets + aggregation data + camera data used by 
  - Choosing either will cause the browser to download a .CSV with headers but no result rows (this is expected as localhost uses staging data).
  - Code works perfectly, but we may need to reconsider the text we use for the buttons. @aliburchard, thoughts?
- As noted by Simone, there is no 'EASYMODE' which translates the raw CSV headers into something more friendly/decipherable to humans; I will need to adapt the 'EASYMODE' feature from MapExplorer for this.

Status: LGTM
Action: Merging

PR #282, opened 2016-08-16
Fetch existing token. Store it in state. Pass it to assignment url.

### LINKED ISSUE #271: Start assignment button doesn't not persist login to wildcam.org
Reason: there's no single sign-on method.
This can be implemented by passing the toke back to wildcam, either in the url or in the request header and then check the token in wildcam code here https://github.com/zooniverse/wildcam-gorongosa/blob/master/app/stores/user-store.coffee#L98

The url approach is quicker: it has to be something that matches the regex here https://github.com/zooniverse/wildcam-gorongosa/blob/master/app/stores/user-store.coffee#L98. Something like `wildcamgorongosa.org/#/classify/assignment-30/access_token=<apikey>`.

simoneduca:
Fix #271

### DIFF
--- a/src/modules/common/actions/login.js
+++ b/src/modules/common/actions/login.js
@@ -18,7 +18,8 @@ export function setLoginUser(user) {
   return (dispatch) => {
     dispatch({
       type: types.SET_LOGIN_USER,
-      user
+      user,
+      token: oauth._tokenDetails,
     });
   };
 }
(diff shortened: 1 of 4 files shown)


shaunanoordin:
- Testing on locahost:3000, OSX+Chrome52, Incognito
- As a Student, I can access Explorer > Assignments to find find a list of Assignments...
  - ...and clicking on "Start Assignment" will take me to OG WG, with an oAuth token and a Assignment ready to load.
- Existing login functionality doesn't seem to be affected by this update:
  - As a logged-in user, I can still log out of Lab, and
  - as an unlogged-in user, I can still log in using a Panoptes ID.

Status: LGTM
Actions: Merging

Issue #276, opened 2016-08-12
CSVs downloaded from assignments page are old format


aliburchard:
Not sure if this falls to @shaunanoordin or @simoneduca but the csvs downloaded from the assignments page use the old headers instead of the ones you get if you download from the map explorer.

aliburchard:
@shaunanoordin we just chatted about this this morning. We should make the aggregated data download as csv easy mode. We need to think about what format the raw classification data (i.e. "my data") should take. I suspect it should be as close to the aggregated data format as possible, but will discuss with Bridget.

simoneduca:
Closed via #286

Issue #278, opened 2016-08-12
Next Milestone Overview


aliburchard:
High Priority items based on testing updates. For after we close the Aug 5 Milestone (which we are about to do!!) 
- [x] login redirect: https://github.com/zooniverse/wildcam-gorongosa-education/issues/271. But #283. 
- [x] **update csv headers for data downloaded on assignments page** - Shaun (see #276) (PR: #286)
- [x] remove beta tag on wildcam lab link
- [x] Remove the first section of text from the landing page of the Lab where it says that this is a work in progress. 
- [x] relabel "tutorial" tab as "help"

Medium High Priority items 
- [x] update progress in real time need @marten for this 
- [x] message to students when they've finished an assignment (in Wildcam OG https://github.com/zooniverse/wildcam-gorongosa/issues/237) (PR: zooniverse/wildcam-gorongosa#245)
- [x] figure out how on earth we have the assignments as truncated by simone talk to subjects as limited in the map explorer. **Not sure this one is still relevant. I'm going to add a warning for teachers when number of images < subjects selected**
- [x] fix safari download bug (PR: #286)
- [x] editing assignments doesn't work

shaunanoordin:
Milestone complete.

PR #286, opened 2016-08-17
Fix 284 by adding AlternateDownloader to handle Safari's nonsense

### LINKED ISSUE #284: Download CSVs don't work properly in Safari
Issue: When choosing to download a CSV on Safari (e.g. using the Download option in Map Explorer), instead of getting the 'SaveAs' dialog box, the window just shows the raw CSV data as a text output.
Analysis:
- Blob output, actually. Trying to Cmd-Save in Safari causes a less than optimal "Save As... _Web Archive_" option instead of a "Save As CSV" option.
- Oh, Safari, you don't like the `download` attribute, do you? `<a href="file.csv" download>Download</a>` http://caniuse.com/#feat=download
  Possible solutions:
- Move the download feature from the app-level to the server level. Good luck!
  Status: Perplexing

shaunanoordin:
Fixes #284 

What?
- When a user (using Safari) attempts to download a CSV, she/he will properly receive a 'Save As' dialog (or saved immediately, depending on user settings) for a .CSV file... as expected.
- Currently this only works with the Map Explorer's Download feature; a solution for Students>Assignments is in progress.

How?
- Since Safari doesn't like SaveAs, `<a href="..." download>` or any app-level solution for forcing a Save As dialog to open, the AlternateDownloader sends the data to the EduAPI, which then echoes the data back as a server response full of `content-disposition: download this goddamn thing as a file` headers.
- This requires a hidden `<form>` that triggers the "old-fashioned" HTTP request that Safari needs.
- Much thanks to @marten for the EduAPI download echo service ( zooniverse/education-api#30 ) that this relies on and to @astopy for increasing the EduAPI's upload limit.

**WARNING:** it may be conceivable that, in the future, the amount of data that a user can query will grow even larger, exceeding the already generous upload limit, so we need to consider preemptively curbing the download size at the app level. Right now, the largest CSV file that can get echoed (uploaded-then-downloaded-again) is about 12MB in size, if the user requests for _all_ aggregated results. I'm looking into zipping the content to reduce strain on the EduAPI.

@simoneduca, please feel free to review and comment on this, as I'm thinking of generalising the CSV download function into a component that can be used by Students>Assignments, not just Map Explorer.

@ali I just realised how ironic it is that it's so difficult to get a Safari going in Gorongosa.


package.json
```diff
@@ -62,6 +62,7 @@
     "express": "^4.13.3",
     "extract-text-webpack-plugin": "^0.9.1",
     "file-loader": "^0.8.5",
+    "gzip-js": "0.3.2",
```
shaunanoordin: `gzip-js` code hasn't been implemented in AlternateDownloader yet; if the zipping solution is ultimately unused, someone please prod me and remind me to remove this.

src/constants/mapExplorer.config.json
```diff
@@ -10,7 +10,7 @@
...
     "sqlQueryViewCameraData": "SELECT DISTINCT location FROM (SELECT cam.latitude, cam.longitude, cam.veg_type, cam.dist_humans_m, cam.human_type, cam.dist_water_m, cam.water_type, sbj.* FROM {CAMERAS} as cam INNER JOIN {SUBJECTS} as sbj ON cam.id = sbj.camera) as camsbj INNER JOIN (SELECT * FROM {AGGREGATIONS} WHERE num_classifications >= 5) as agg ON camsbj.subject_id = agg.subject_id {WHERE}",
     "sqlQueryViewCameraMeta": "SELECT * FROM {CAMERAS} {WHERE}",
-    "sqlQueryDownload": "SELECT cameras.*, items.* FROM {CAMERAS} AS cameras LEFT JOIN (SELECT a.camera, a.location, a.dateutc, a.month, a.year, a.season, a.time_period, b.* FROM {SUBJECTS} as a INNER JOIN (SELECT * FROM {AGGREGATIONS} WHERE num_classifications >= 5) as b ON a.subject_id = b.subject_id) AS items ON cameras.id = items.camera {WHERE} LIMIT 10000",
+    "sqlQueryDownload": "SELECT cameras.*, items.* FROM {CAMERAS} AS cameras LEFT JOIN (SELECT a.camera, a.location, a.dateutc, a.month, a.year, a.season, a.time_period, b.* FROM {SUBJECTS} as a INNER JOIN (SELECT * FROM {AGGREGATIONS} WHERE num_classifications >= 5) as b ON a.subject_id = b.subject_id) AS items ON cameras.id = items.camera {WHERE}",
```
shaunanoordin, line 13: NO MORE LIMITS. NO MORE LIMITS.

shaunanoordin:
Just to clarify the New Plan:
- I'm creating the SuperDownloadButton component that handles all CartoDB-download functions.
- Once functional (i.e. it stops bugging out on me), I'll slot it into Students > Assignments
- This will also implement the CSV "EASYMODE" header translations to Students > Assignments, and will fix #276 and address " update csv headers for data downloaded on assignments page" in #278.

I'm taking a break from the zipping functionality for the moment; I may need to simply limit all downloads to 10000 items like before for the time being, while I figure out wtf is up with the file formats I'm getting back.

The individual functions are working fine, but since they tend to be rather interlocked in the supercomponents they exist in, I'm still trying to find an elegant way to package all those functions into a standalone SuperDownloadButton component.

shaunanoordin:
- The "Download from CartoDB" functionality that was spread across Map Explorer and Students Assignment have been consolidated into a single SuperDownloadButton component.
- EASYMODE is built in, but has not been enabled for Student Assignments yet - I'm still looking at the best CSV format for those CSVs

shaunanoordin:
- EASYMODE enabled! Download buttons on Map Explorer and Student Assignments now provide a CSV that's somewhat more readable to humans... and are all standardised to one format, to boot.
- I'll need to test this with live data (`zootester1` on production after doing a few Classifications on an Assignment) to know for sure, but...
  - ...I suspect I may need to get another PR to tweak the difference between percentage_behaviour (aggregated results) and what_i_marked_as_behaviour (My Classifications) values.

screen shot 2016-08-19 at 17 55 20
_New look_

@simoneduca, this is ready for review when we're back next week! Woot!

simoneduca:
On my screen 

screen shot 2016-08-22 at 14 50 30

I'd recommend setting `max-width: 75em` in `.student-assignmentlist`

simoneduca:
Please add `readOnly` to the input fields in the form of `SuperDownloadButton` to suppress react warning `Warning: Failed form propType: You provided a`value`prop to a form field without an`onChange`handler. This will render a read-only field. If the field should be mutable use`defaultValue`. Otherwise, set either`onChange`or`readOnly`. Check the render method of`SuperDownloadButton`.`

shaunanoordin:
Update:
- aria-xxx and readOnly tags added to SuperDownloadButton to take care of errors.
- Action Buttons in Student Assignment List rearranged to stack vertically, to better support content when Assignments have long descriptions.

screen shot 2016-08-22 at 16 23 09

👍 Thanks for the feedback, @simoneduca!

simoneduca:
LGTM 👍

PR #287, opened 2016-08-18
Add warning on submit.

simoneduca:


shaunanoordin:
- Tested on localhost:3000+OSX+Chrome52
- As a Teacher, if I try to create an Assignment, then use the Map Explorer to select 6 subjects, then I will get an error message if I set "Number of Images" in the New Assignment form to 7 or more.
  - The New Assignment form only works if the Number of Images is set to be less or equals to the number of subjects selected.
  - (Note on testing on localhost: there will only be 6 subjects selected, due to hardcoding.)

I vaguely feel that the teachers may get somewhat confused by the wording of "The number of subjects selected can't be smaller than the number of images.", but this is because of the differences in terminology between us and general users.

I'd really like to suggest an improvement, but e.g. "The Number of Images Each Student Needs to Classify must not exceed the number of Images you selected from the map" may be equally confusing.

The code itself works, though, so I'm keen to merge & see how actual users react to the Lab.

Status: LGTM
Actions: Merging

PR #297, opened 2016-08-23
Small text changes.

### LINKED ISSUE #296: minor text edits
**Data Page:**

When you click “Select for Assignment”, in the popup box, change to:

We've found X results.

How many would you like to use for the Assignment you're creating?

Tip: The images you assign to your students will be selected from this set. If you want all of your students to identify the same images, make this number the same as the number you assigned per student. If you want each student to identify different images from this set, make this number large enough to provide different images for all of your students.

**Assignments Page:**

Change the buttons to “My Data” and “Public Data”

simoneduca:
Fix #296


shaunanoordin:
- Text updates confirmd on Teacher > Map Explorer

Status: LGTM
Actions: Merging

PR #301, opened 2016-09-21
Allow users to (partially) edit Assignments

### LINKED ISSUE #300: Can we View/Edit Assignments?
@simoneduca I need some refreshers on this - do we allow users to Edit Assignments?

I received an e-mail from Milena Tsvetkova of the OII with the following query/request: _"I further have another question — is it intentional that I cannot edit the students and the subject selection once the assignment is created and I click on the “View/Edit” action? Any chance you can make these fields editable so that I don’t have to create a new assignment if I mess it up."_

If we're making users unable to edit Assignments on purpose, I'll need to change the "View/Edit" button to just "View". If users are suppose to be able to Edit Assignments, then a larger fix is needed as it seems non-functional at the moment.

shaunanoordin:
## Overview

Fixes #300 

With this PR, Teachers can now successfully Edit Assignments. However, they can only change the following parameters of an existing Assignment:
- Title
- Description
- Due Date

The following parameters are **locked**:
- Subjects
- Students
- Images per student
## Notes

As stated in #300, you won't see the "cannot edit Assignment" error on the current version of `master` on `localhost`.
## Status

Ready for review

### DIFF
--- a/src/modules/teachers/components/AssignmentForm.jsx
+++ b/src/modules/teachers/components/AssignmentForm.jsx
@@ -96,7 +96,7 @@ class AssignmentForm extends Component {
     //WildCam Gorongosa has no Staging data that we can use to create Subjects.
     //Therefore, we need to hardcode some Subject IDs.
     //--------
-    if (window.location.hostname === 'localhost') {
+    if (!this.editMode() && window.location.hostname === 'localhost') {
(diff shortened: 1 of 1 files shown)


simoneduca (approved):
LGTM

PR #302, opened 2016-09-27
Removed old MapExplorer components

shaunanoordin:
## Overview

The old version of MapExplorer (v2) has been removed from the repo. It was only there for safety/transitioning purposes and it's no longer needed.

The current/latest version, v3, has been left untouched.
## Status

Ready for review


simoneduca (approved):
Noce one, lgtm.

PR #304, opened 2016-10-05
Added "Easy Mode Special Values" for CSVs

### LINKED ISSUE #303: CSV species count has non-numbers: "11-50" and "50+"
## Overview

When a Teacher/Student downloads a CSV (from either the Map Explorer or the Assignments page), the column for "number" (i.e. aggregated species count) can have the _numeric range/text_ "11-50" and "50+" in addition to the numbers 1 to 10.

For some users, this is an issue because they expect specific _integers_ in their CSV data, not a numeric range or text.
## Notes

This issue was originally reported by Bridget in an e-mail to me on 29 Sep 2016. Of note, the symptoms she described was:

> Sorry to bother you again, but another teacher just brought another issue to my attention. In the csv download (see the attached file I downloaded today), if you look at the species_count column and scroll down to row 148, you’ll see that the species count for that image shows up as “Nov-50” instead of a number. This same thing appears in several other rows further down. If you convert the entire column to a “number” format, the date changes from “Nov-50” to 17106.

In this scenario, the app she used (presumably Excel) to open the CSV misinterpreted the _numeric range_ "11-50" as the _date_ November the 50th. This illustrates one example of how the mix of numbers and numeric ranges/text in a single column could make things difficult for teachers and users.
## Possible Solution

The most straightforward solution would be one of the following:
- "Massage" the CSV data so "11-50" is translated to "11 to 50" and "50+" translated to "50 or more", or
- "Massage" the CSV data so "11-50" is translated to "25" (mean/estimate) and "50+" translated to "50" (min/estimate), or
- Create a message informing Teachers/Students that "you might see values like 11-50 and 50+ in your data, which simply indicates an unexpectedly(?) large amount of animals"

**However,** we need a subject expert to weigh in on this, as there are two conflicitng goals here: between usability (Teachers/Students obviously cannot use an ambiguous range like "11-50" and "50+" in standard numerical calculations and charts) and scientific accuracy (the range 11-50 isn't the same as the mean/estimate of 25).

(As I understand it, the WildCam Classification/Survey task was designed this way for a reason, as excessively large discrete counts (i.e. actually putting in ninety integers from 11 to 100) would not have contributed to the research or survey efforts.)

@aliburchard do you have any thoughts?

shaunanoordin:
## Overview

Fixes #303 

When downloading CSVs, the values "11-50" and "51+" (i.e. very large species counts) are considered special values. They'll be translated to 25 and 75 respectively. This is to allow Teachers and Students (particularly those using Excel) to use "proper" numerical values to, e.g. create charts and graphs
## Status

In Progress.

Need to add Caveats and Warnings for users.

### DIFF
--- a/src/modules/common/components/SuperDownloadButton.jsx
+++ b/src/modules/common/components/SuperDownloadButton.jsx
@@ -118,7 +118,8 @@ class SuperDownloadButton extends Component {
     //Prepare the column headers
     //NOTE: we can package these translations into the SQL query but it'll get super long & messy
     for (let key in mapconfig.cartodb.csvEasyModeTranslator) {
-      row.push('"'+mapconfig.cartodb.csvEasyModeTranslator[key].replace(/"/g, '\\"')+'"');
+      let thisVal = '"'+mapconfig.cartodb.csvEasyModeTranslator[key].replace(/"/g, '\\"')+'"'
+      row.push(thisVal);
     }
     row = row.join(',');
     data.push(row);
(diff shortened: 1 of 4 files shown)


shaunanoordin:
## Status (Update)

Ready for review

src/modules/common/components/SuperDownloadButton.jsx
```diff
@@ -127,9 +128,15 @@ class SuperDownloadButton extends Component {
     json.rows.map((rowItem) => {
       let row = [];
       for (let key in mapconfig.cartodb.csvEasyModeTranslator) {
-        (json.fields[key].type === 'string' && rowItem[key])
-          ? row.push('"'+rowItem[key].replace(/"/g, '\\"')+'"')
-          : row.push(rowItem[key]);
+        let thisVal = rowItem[key];
+        
+        if (mapconfig.cartodb.csvEasyModeSpecialValues[thisVal]) {
+          //Teacher & Students using Excel are confused when seeing the range "11-50" and "51+"; these speical values will be translated to "25" and "75"
```
simoneduca, line 134: "special"

simoneduca (approved):

PR #307, opened 2016-11-15
Prevents entering a negative number of subjects.

simoneduca:
It works also if user tries to type it manually.


shaunanoordin:
## PR Review
- When creating Assignments, negative numbers can no longer be set for "number of images"
- Tested on Firefox49/Chrome54+OSX

## Status

Merging and deploying

simoneduca:
Made a pig's ear of this, but I added the check for type here.

PR #309, opened 2016-11-15
Remove dash only if input type is number.

simoneduca:


shaunanoordin:
## PR Review
- Replaces #307 (and #308)
- When creating Assignments, negative numbers can no longer be set for "number of images"
- However, unlike #307, we can still use the dash (`-`) when creating Assignment titles and descriptions.
- Tests OK on Chrome54+OSX

## Status

Merging and deploying

PR #313, opened 2017-01-11
Update Panoptes application ID

marten:


shaunanoordin:
## PR Review
* TL;DR we have a new application ID. Merging and deploying immediately.

PR #323, opened 2017-11-15
Add WG program

srallen:
This adds the notion of a program to the redux store including the associated actions, reducer, etc. I've added the request for the wildcam gorongosa program to the App component and updated the config to work with staging. 

Follow up to this is to add the program filter to the get request for classrooms and the program link on create.


shaunanoordin:
Ooh, nice, WildCam Gorongosa Lab's getting ready for the future!

Right, just give me a minute to debug something - it looks like the REQUEST_PROGRAM action makes a call to `http://localhost:3000/undefinedprograms/undefined`, which should be straightforward enough to diagnose.

shaunanoordin:
/giphy eyeroll

Of course that derped - the WildCam Lab is using a very old implementation of config.js, one that's
1. not smart enough to listen for `?env=` modifiers to switch to the correct environment, and
2. doesn't recognise that `staging === development` for config environments.

brb, this will be fixed.

srallen:
FWIW, programs don't exist yet on production and that's what we're planning on deploying when we're all together, so it'll still error when using the production env. However, this does make it easier for us to test production before deploying when the time comes.

shaunanoordin (approved):
## PR Review
This PR prepares WildCam Gorongosa Lab for the eventual upgrade of the EduAPI, which requires that each unique _education project_ (Gorongosa, Darien, I2A, etc) have their own unique "Program" ID.

- This PR fetches Program data from the EduAPI,
  - but does not do anything with it yet. (NOTE: In the future, App.jsx may need to stop the whole website from displaying content until the project data is fully loaded - an "App loading..." screen, if you will)
  - but only has a Program ID for staging at the moment; a production counterpart will be coming later.
- Side improvement: the config has been updated so that:
  - it acknowledges the 'development' environment (as the same as the 'staging' environment)
  - it acknowledges the `?env` switch.

### Status
LGTM! The project works on localhost and acknowledges ?env switches correctly; I can also confirm that the call to the Edu API's 'programs' endpoint works correctly (on staging/development)

Approved, merging, go!

PR #325, opened 2017-11-29
Program filter and assignment create update

srallen:
This updates the WG:L for the current API:

- Classroom GET requests are now filtered by program
- Links the classroom to the WG program on classroom create
- Assignment creation uses a workflow id
  - In the staging environment, the assignment creation uses WG project on staging and a hard coded set of subject ids. This is because the Carto DB data is only for production.

The assignment creation still isn't fully testable because as noted on slack, there's still an issue with the POST request failing permissions with Panoptes, however, the code itself could be reviewed in the interim.

### DIFF
--- a/src/modules/teachers/actions/assignment.js
+++ b/src/modules/teachers/actions/assignment.js
@@ -7,7 +7,7 @@ import * as types from '../../../constants/actionTypes';
 
 
 // Action creators
-const { root, assignments } = config.eduAPI;
+const { root, assignments, workflowId } = config.eduAPI;
 
 export function createAssignment(assignment, classroomId) {
   const classroomData = {
@@ -29,12 +29,13 @@ export function createAssignment(assignment, classroomId) {
     id: subject_id,
     type: 'subjects',
   }));
-  
+
   const bodyData = JSON.stringify({
     data: {
       attributes: {
         name: assignment.name,
         metadata,
+        workflow_id: workflowId
       },
       relationships: {
         classroom: {
@@ -109,12 +110,12 @@ export function deleteAssignment(assignmentId, classroomId) {
 }
 
 export function editAssignment(fields, assignment) {
-  
+
   let studentData = fields.students.map(student_id => ({
     id: student_id,
     type: 'student_user',
   }));
-  
+
   const bodyData = JSON.stringify({
     data: {
       attributes: {
@@ -134,7 +135,7 @@ export function editAssignment(fields, assignment) {
       },
     }
   });
-  
+
   return dispatch => {
     dispatch({
       ...fields,
--- a/src/modules/teachers/actions/assignment.js
+++ b/src/modules/teachers/actions/assignment.js
@@ -2,7 +2,7 @@ import { browserHistory } from 'react-router';
 import fetch from 'isomorphic-fetch';
 import apiClient from 'panoptes-client/lib/api-client';
 
-import config from '../../../constants/config';
+import config, { env } from '../../../constants/config';
 import * as types from '../../../constants/actionTypes';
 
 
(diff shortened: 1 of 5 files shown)


srallen:
@shaunanoordin I think this is ready for review. One last change is that I added the program filter to the student classroom request. The assignment POST works now with a successful 201 response, though, there's still one last API side bug with the new subject set getting the subjects in the POST linked. So far that appears to be an API only bug.

src/modules/teachers/actions/assignment.js
```diff
@@ -7,34 +7,55 @@ import * as types from '../../../constants/actionTypes';
...
     type: 'student_user',
   }));
-  const subjectData = assignment.subjects.map(subject_id => ({
-    id: subject_id,
-    type: 'subjects',
-  }));
-  
+
+  // Carto DB doesn't have staging data. We use hard coded subject ids from the staging WG project and ignore the Carto selection.
+  if (process.env.NODE_ENV === 'staging') {
```
shaunanoordin: Should this check using the `env` variable from `config.js` instead of using process.env.NODE_ENV?

shaunanoordin: Also, should it check for both `env === 'staging' || env === 'development'` ?

srallen:
Noting that the subjects linking now works on the API.

shaunanoordin (approved):
## PR Review
This PR adds the newly-required Program ID to requests to the Edu API.

ALSO:
- Assignment creation in staging/development now uses a pre-made list of Subjects call the Sample Subjects.

Context:
- The early Edu API was designed to only work with the WildCam Gorongosa "program".
- With the recent introduction of the I2A and WildCam Darien "programs" to the EduAPI, it's important that we have a way to, e.g., differentiate which Program's Classrooms we're asking for when we query the API.
- Hence, the `?program_id` query parameter/filter.

Affected:
- Requesting Classrooms for Teachers (tests OK)
- Requesting Classrooms for Students

External dependency:
- Edu API. The staging version has already been updated so that _it requires_ the `?program_id` parameter to be specified.
  - NOTE: do NOT deploy latest changes to production until we can confirm that the production version of Edu API also requires the `?program_id`
- WildCam Gorongosa _staging project setup on Panoptes-staging._
  - We need a sorta 'actual' Zooniverse project set up in Panoptes-staging to reflect our WildCam Gorongosa "program" on EduAPI-staging.
  - This is because EduAPI-staging will check for valid Subject IDs and Workflow IDs when creating Assignments.

### Status
Code LGTM. Tests look good too, since - as a Teacher (`zootester1`) on localhost - I'm able to create Classrooms and Assignments just fine.

Approved, merging.

shaunanoordin:
I just remembered one thing that'll make things a bit weird: there's already a previous +"hardcode the selected Subjects for Assignments when on staging/development",_ code, hidden in `AssignmentForm.jsx`. It does exactly what the new code does, and it looks like...

```
    //TESTING: localhost uses 'Bacon' Zooniverse project for Subject IDs.
    //WildCam Gorongosa has no Staging data that we can use to create Subjects.
    //Therefore, we need to hardcode some Subject IDs.
    //--------
    if (!this.editMode() && window.location.hostname === 'localhost') {
      newAssignment.subjects = [
          '4077',
          '4079',
          '4078',
          '4076',
          '4080',
          '4075',
        ];
    }
    //--------
```

OK, a few notes for this one:
- The subjects needed to be added at `AssignmentForm.handleSubmit()` (i.e. `newAssignment.subjects = [...]`) because otherwise the AssignmentForm wouldn't validate.
- The Bacon Subjects were chosen because... back then, EduAPI just needed valid (on staging) Subject IDs, not their associated workflows or whatnot.
- This old piece of code doesn't even check on the `env` variable, just... just localhost, wtf.

As a result, I'm marking the Bacon code as something that needs to be fixed _separately_ from this PR.

It doesn't matter in practice, since the duplication just means that...
- the old "hardcoded subjects" code in `AssignmentForm.jsx` allows the Create/Edit Assignment form to be validated & submitted _on staging/development_
- the new "hardcoded subjects" code in `assignment.js` allows Classroom API calls to work _on staging/development_

